PR #56, opened 2020-05-29
[RFC] build lxd binaries without using go plugin's build step

cjp256:
@sergiusens mentioned earlier that we effectively broke the customized go plugin by changing some internal interfaces.  I took a quick look and thought maybe we should do away with the custom plugin and stop relying on the go plugin's build step.  The override-build is already heavily customized...

This hasn't been thoroughly tested yet, and could be cleaned up a bit, but wanted to share and get your thoughts @stgraber.

---

Perform the go build commands performed by the go plugin in
`override-build`.  The `override-build` is already very heavily
customized and there is little need to rely on the plugin anymore.

This patch simply adds a few extra build commands with the
correctly exported CGO_LDFLAGS to find the library dependencies
that are being built/staged.

Remove the custom go plugin, and rely on the `go` plugin again
but only using its `pull` functionality.


stgraber:
The pull part is why we had to come up with `nogetgo`, that's literally the only difference between the stock Go plugin and ours :)

The Go plugin runs `go get` on the source. If the source fails to `go get`, then the snap build fails.
In our case we need to apply a whole load of git cherry-picks between the actual pull (as in `git pull`) and the `go get` step.

That's what our fork of the go plugin allows us to do but just skipping this one bit.

stgraber:
I would expect your change to work fine on edge as there are no cherry-picks there, but it will re-introduce the issue that `nogetgo` fixes on any of the other branches.

PR #67, opened 2020-09-15
buginfo: Use "--format=yaml" in list commands

arossbell:
lxc.buginfo has a few "lxc x list" commands listed without any specified
format.  This defaults to the "table" format.  However, the default "table"
format does not provide some information that another format, like the "yaml"
format, provides.  Some of the additional information could be useful for
engineers debugging a lxc setup.  This PR appends "--format=yaml" to the
applicable list commands.

Signed-off-by: Adam Bell <adam.bell@canonical.com>

### DIFF
--- a/snapcraft/commands/buginfo
+++ b/snapcraft/commands/buginfo
@@ -45,37 +45,37 @@ if lxc info >/dev/null 2>&1; then
 
     echo "## Instances"
     echo '```'
-    lxc list
+    lxc list --format=yaml
     echo '```'
     echo ""
 
     echo "## Images"
     echo '```'
-    lxc image list
+    lxc image list --format=yaml
     echo '```'
     echo ""
 
     echo "## Storage pools"
     echo '```'
-    lxc storage list
+    lxc storage list --format=yaml
     echo '```'
     echo ""
 
     echo "## Networks"
     echo '```'
-    lxc network list
+    lxc network list --format=yaml
     echo '```'
     echo ""
 
     echo "## Projects"
     echo '```'
-    lxc project list
+    lxc project list --format=yaml
     echo '```'
     echo ""
 
     echo "## Profiles"
     echo '```'
-    lxc profile list
+    lxc profile list --format=yaml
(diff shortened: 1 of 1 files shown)


stgraber:
The YAML output tends to be quite painful to read though as it contains a LOT more information.

So this would make it significantly harder for us (upstream) to read a bug report made from a user, not to mention that a busy system with a few hundreds containers each with 50 or so snapshots would make the output so large that it would overflow Github and most pastebins.

If there's a specific bit of information that's missing and that's constantly requested, then we should add that, but we should not switch everything over to YAML as that will actually make it harder for us to process.

arossbell:
That's fair -- thank you!

arossbell:
@stgraber If there was a flag added to this command (ie "--long-info", "--format=yaml", or "--verbose") that could give the full list, would that be acceptable?
Right now, the sosreport tool that our support team uses with customers [is using the lxc.buginfo command](https://github.com/sosreport/sos/blob/master/sos/report/plugins/lxd.py) to collect LXC-related information (if LXC is installed from a snap). We receive those in compressed archives -- in that sense, it wouldn't be a concern there about overflowing where it's going (but overloading an engineer is still possible 🙂).

If you think it would be acceptable to add a flag to allow a much more detailed output, I could put that together on this commit + PR. That can leave the regular `lxc.buginfo` output unchanged while giving an easy way for the additional information to be collected. Any thoughts on that?

stgraber:
Yeah, having a `--verbose` or similar flag which dumps more info should be fine, just beware that if you forward such a report to us, there's a good chance we'll either take a while before we get to it or we'll just ask for a clean `lxd.buginfo` run :)

arossbell:
To wrap this up: I wrote up some internal support documentation instead (for the time). If we start getting more LXC/LXD-specific issues, I/we may revisit something like this, but for the time, the documentation will *probably* be sufficient.

You guys write so good of code that customers don't break it. :slightly_smiling_face: 

Thanks again!

PR #78, opened 2021-07-06
Drop support for ZFS 0.6 as Ubuntu 16.04 is EOL

simondeziel:


stgraber:
Not going to be merging that as we still have over 100k active users on 16.04 which this would break :)

stgraber:
@simondeziel we also don't want to break lxd-migrate as the safest upgrade method out of 16.04 is to switch to the snap first, update LXD and then upgrade the OS.

The escape change for ZFS sounds good though :)

PR #86, opened 2021-09-21
Don't try to install edk2 deps on non-edk2 architectures

wgrant:
Some of them don't exist everywhere, e.g. acpica-tools isn't available
on riscv64 in focal. override-build short-circuits outside amd64 and
arm64, so don't require the build-deps either.

It seems like the dpkg features of architecture lists and negative
matches weren't carried over, so there's no non-duplicatastic way to do
this AFAICS.


stgraber:
Thanks!

PR #94, opened 2022-03-18
snapcraft.yaml: pull meson early

xnox:
Plugins cannot pull dependenices during pull step. Instead, add a part that pulls meson during pull step. This helps with completing build steps offline.

I recently managed to complete LXD riscv64 build in Launchpad. This patch was used during that successful build, please consider applying it.


stgraber:
Tweaked the diff a tiny bit to make it clearer what's going on here.

PR #96, opened 2022-03-18
snapcraft.yaml: go get during pull stage, instead of build

xnox:
Ensure that go get is done during pull, instead of build step.

I am not a regular golang coder, this pull request needs especially careful review to ensure that go-get & go-build actually get build-deps which are then used by go-build, instead of like resulting in duplicate downloads that are not reused.

I recently managed to complete LXD riscv64 build in Launchpad. This patch was used during that successful build, please consider applying it.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1213,6 +1213,20 @@ parts:
       - usbutils
       - xdelta3
     plugin: nil
+    override-pull: |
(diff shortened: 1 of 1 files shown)


stgraber:
Fixed a few small issues with this and cleaned up the diff a bit.

PR #97, opened 2022-03-18
snapcraft.yaml: use core22-step-dependencies

xnox:
This build attribute insures that all parts are pulled during LP's snapcraft pull call, and that they are not subsequently repulled again during build.

This requires for pull steps of each part to not require the before parts to be pulled.

This feature is currently present in snapcraft latest/edge only, possibly will be available in 7.x/6.0.3.

This means this patch cannot be applied, unless the builds are changed to use snapcraft from latest/edge which is undesirable.

I recently managed to complete LXD riscv64 build in Launchpad. This patch was used during that successful build. After establishing 6h build time, the network proxy allowed time has been requested to be increased. So it might be the case that lxd snap riscv64 builds might be successful now even without this patch applied.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -146,6 +146,7 @@ hooks:
 parts:
   # Dependencies
   btrfs:
+    build-attributes: [core22-step-dependencies]
     source: snapcraft/empty
     plugin: nil
     stage-packages:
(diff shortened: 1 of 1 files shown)


xnox:
This can be merged and used once snapcraft 6.1 is out

https://forum.snapcraft.io/t/call-for-testing-snapcraft-6-1/29210

https://github.com/snapcore/snapcraft/releases/tag/6.1

PR #99, opened 2022-03-31
lxcfs: fix nesting setup for backward compatibility purpose

adglkh:
For backward compatibility, hand `/var/lib/lxcfs` through the host to
the container being as the lxcfs mount point.

Signed-off-by: gary-wzl77 <gary.wang@canonical.com>

### DIFF
--- a/patches/lxcfs-0001-hook.patch
+++ b/patches/lxcfs-0001-hook.patch
@@ -1,6 +1,6 @@
---- /root/parts/lxcfs/install/snap/lxd/current/lxcfs/lxc.mount.hook	2022-03-23 22:47:13.153225112 +0000
-+++ lxc.mount.hook	2022-03-23 23:23:21.614527054 +0000
-@@ -31,8 +31,10 @@ if [ -d /var/snap/lxd/common/var/lib/lxc
+--- /root/parts/lxcfs/install/snap/lxd/current/lxcfs/lxc.mount.hook	2022-03-31 11:35:47.937752513 +0800
++++ lxc.mount.hook	2022-03-31 11:30:15.094738379 +0800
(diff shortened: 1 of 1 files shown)


patches/lxcfs-0001-hook.patch
```diff
@@ -10,6 +10,12 @@
 +    rm -Rf "${LXC_ROOTFS_MOUNT}/var/snap/lxd/common/var/lib/lxcfs"
 +    mkdir -p "${LXC_ROOTFS_MOUNT}/var/snap/lxd/common/var/lib/lxcfs"
 +    mount -n --bind /var/snap/lxd/common/var/lib/lxcfs "${LXC_ROOTFS_MOUNT}/var/snap/lxd/common/var/lib/lxcfs/"
++fi
++
++# For backward compatibility, hand `/var/lib/lxcfs` through the host to
++# the container being as the lxcfs mount point.
++if [ -d ${LXC_ROOTFS_MOUNT}/var/lib/lxcfs/ ]; then
++    mount -n --bind /var/snap/lxd/common/var/lib/lxcfs ${LXC_ROOTFS_MOUNT}/var/lib/lxcfs/
```
stgraber: Can you make the path quoting match what we have for the hunk above?
Something like:
```
if [ -d "${LXC_ROOTFS_MOUNT}/var/lib/lxcfs/" ]; then
    mount -n --bind /var/snap/lxd/common/var/lib/lxcfs "${LXC_ROOTFS_MOUNT}/var/lib/lxcfs/"
fi
```

stgraber: @adglkh

adglkh: Thanks. fixed.

PR #101, opened 2022-04-19
Allow usage of external OVS tools.

VincentK-Titandc:
`ovs-appctl` and `ovs-vsctl` tools can be linked to host FS on-demand as
some users may want to use their own version of these tools.

_Note_: `ovs-appctl` was mainly added to override access to its embedded
version as it seems that LXD is not using it anyway.

Signed-off-by: Vincent KHERBACHE <v.kherbache@titandc.fr>


stgraber:
Looks good, thanks!

PR #102, opened 2022-04-26
Change systemd unit arguments from hard-coded lxd to SNAP_INSTANCE_NAME.

virtuous-sloth:
Creating your own personal snap of lxd with a different name fails during installation because the daemon.activate and daemon.start scripts hard-code the snap name to lxd:
```
# sudo snap install --devmode virtuous-sloth-lxd_0+git.d6919a5-dirty_amd64.snap
- Start snap "virtuous-sloth-lxd" (unset) services (systemctl command [start snap.virtuous-sloth-lxd.activate.service] failed with exit status 1: Job for snap.virtuous-sloth-lxd.activate.service failed because the control process exited with error code.
See "systemctl status snap.virtuous-sloth-lxd.activate.service" and "journalctl -xeu snap.virtuous-sloth-lxd.activate.service" for details.
```
I can do devmode without changing the snap name but would like to try pusblishing virtuous-sloth-lxd in snapcraft.io for my own edification of the whole process.

I chose SNAP_INSTANCE_NAME instead of SNAP_NAME even though lxd does not support multiple instances because it seemed the right thing to do anyway.

### DIFF
--- a/snapcraft/commands/daemon.activate
+++ b/snapcraft/commands/daemon.activate
@@ -39,11 +39,11 @@ daemon_user_group="${daemon_user_group:-"lxd"}"
 
 # Detect missing socket activation support
 echo "==> Checking for socket activation support"
-if ! nsenter -t 1 -m systemctl is-active -q snap.lxd.daemon.unix.socket; then
+if ! nsenter -t 1 -m systemctl is-active -q snap.${SNAP_INSTANCE_NAME}.daemon.unix.socket; then
     sleep 3s
-    if ! nsenter -t 1 -m systemctl is-active -q snap.lxd.daemon.unix.socket; then
+    if ! nsenter -t 1 -m systemctl is-active -q snap.${SNAP_INSTANCE_NAME}.daemon.unix.socket; then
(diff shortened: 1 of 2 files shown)


nicolasbock (commented):
If you rename the snap, wouldn't there be more places where you need to change `lxd` -> `${SNAP_INSTANCE_NAME}`, e.g.

https://github.com/lxc/lxd-pkg-snap/blob/12fc2d12b4eca9c38764df1c26230ff0e6830f77/snapcraft/etc/logrotate.conf#L1

It would also help future investigators of this code base if your pull request description was in the commit message. Remember that the pull request is not part of the git repository and lives only on GitHub.

stgraber:
Branch needs a rebase, it currently shows 3 commits when it likely should just show one or two

virtuous-sloth:
I *think* that fixed it; it should be just one simple commit with an expanded commit message. 

Please let me know if that is not the case.

nicolasbock:
Yes, now it's only 1 commit 😄

stgraber:
Thanks!

PR #103, opened 2022-07-22
snapcraft: Add support for sideloaded lxd-agent debug executable

tomponline:
Signed-off-by: Thomas Parrott <tomp@tomp.uk>

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -189,6 +189,12 @@ if [ "${lvm_external:-"false"}" = "true" ]; then
     ln -s "${SNAP}/wrappers/run-host" "/run/bin/lvresize"
 fi
 
+# Detect presence of sideloaded lxd-agent executable.
+if [ -x "${SNAP_COMMON}/lxd-agent.debug" ]; then
+    echo "==> WARNING: Using a custom debug lxd-agent binary!"
+    ln -s "${SNAP_COMMON}/lxd-agent.debug" "/run/bin/lxd-agent"
+fi
+
 # Redirect getent to the host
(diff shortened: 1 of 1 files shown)


snapcraft/commands/daemon.start
```diff
@@ -189,6 +189,12 @@ if [ "${lvm_external:-"false"}" = "true" ]; then
     ln -s "${SNAP}/wrappers/run-host" "/run/bin/lvresize"
 fi
 
+# Detect presence of sideloaded lxd-agent executable.
+if [ -x "/var/snap/lxd/common/lxd-agent.debug" ]; then
```
stgraber: This should be `"${SNAP_COMMON}/lxd-agent.debug"`

snapcraft/commands/daemon.start
```diff
@@ -189,6 +189,12 @@ if [ "${lvm_external:-"false"}" = "true" ]; then
...
+    echo "==> WARNING: Using a custom debug lxd-agent binary!"
+    ln -s "/var/snap/lxd/common/lxd-agent.debug" "/run/bin/lxd-agent"
```
stgraber: same here

snapcraft/commands/daemon.start
tomponline: OK cool. I checked initially, and saw quite a few places that reference `/var/snap/lxd/common/` in that file so thought that was the way to do it.

snapcraft/commands/daemon.start
tomponline: Fixed

snapcraft/commands/daemon.start
tomponline: Fixed

PR #108, opened 2022-11-02
Tiny fixes

simondeziel:


snapcraft/commands/lxd
```diff
@@ -24,10 +24,6 @@ export LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-}:${SNAP_CURRENT}/lib:${SNAP_CURRENT}
 export PATH="${PATH}:${SNAP_CURRENT}/bin"
 export LXD_DIR=${LXD_DIR:-"${SNAP_COMMON}/lxd/"}
 
-# Make sure we have a ZFS binary on the path
-export LD_LIBRARY_PATH="${SNAP_CURRENT}/zfs-0.7/lib/:${LD_LIBRARY_PATH}"
-export PATH="${SNAP_CURRENT}/zfs-0.7/bin:${PATH}"
-
```
stgraber: Why is it unneeded?

stgraber: We need a functional `zfs` in PATH when running `lxd init` for it to work properly.

simondeziel: I am assuming that `daemon.start` already took care of adding the proper ZFS version path to those variables. I don't see why that `0.7` version would need special treatment?

stgraber: `daemon.start` and `lxd` are two completely different wrappers.
`daemon.start` is used for the daemon, `lxd` is used for when the user runs the `lxd` command such as `lxd init`.

We need a working version of zfs (doesn't matter which one, which is why we use 0.7) to be on the PATH when someone runs `lxd init`.

Issue #65, opened 2020-09-11
Bump kernel.keys.maxkeys by default


nobuto-m:
As per the document, https://linuxcontainers.org/lxd/docs/master/production-setup, there are some values recommended to bump for the production setup.

At this moment, lxd explicitly bumps fs.inotify.max_user_instances out of the box since it was a common limit we hit previously.
https://bugs.launchpad.net/ubuntu/+source/lxd/+bug/1602192
https://github.com/lxc/lxd-pkg-snap/blob/7c0719817b047d9670f2c4969fdcbb0a01a304f5/snapcraft/commands/daemon.start#L336-L357

Would be nice to add `kernel.keys.maxkeys` into the default list as we see hitting to the limit commonly now.
https://launchpad.net/bugs/1891223

PR #109, opened 2022-12-08
Bump /proc/sys/kernel/keys/maxbytes out of the box

nobuto-m:
The default limit as 20000 can prevent ~20 LXD containers on a host from starting with Ubuntu guest (and cloud-init, snapd and jujud in it). 20 is not a huge number like 100 or 1,000 so it's fair to accommodate it out of the box.

Let's bump it in a similar way to /proc/sys/kernel/keys/maxkeys which we already did some time ago. See more details in
https://launchpad.net/bugs/1891223

Follow-up of #65 and #66

Signed-off-by: Nobuto Murata <nobuto.murata@canonical.com>

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -409,6 +409,13 @@ if [ "$(stat -c '%u' /proc)" = 0 ]; then
         fi
     fi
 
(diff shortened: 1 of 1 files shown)


tomponline (approved):

tomponline:
@stgraber happy for this to be merged?

PR #111, opened 2023-01-11
daemon.start: add lxcfs.extra.opts config option

mihalicyn:
It may be convenient for debugging:
snap set lxd lxcfs.extra.opts="-d"
snap restart lxd


snapcraft/hooks/configure
```diff
@@ -88,6 +89,7 @@ config="${SNAP_COMMON}/config"
     echo "lxcfs_loadavg=${lxcfs_loadavg:-"false"}"
     echo "lxcfs_pidfd=${lxcfs_pidfd:-"false"}"
     echo "lxcfs_cfs=${lxcfs_cfs:-"false"}"
+    echo "lxcfs_extra_opts=\"${lxcfs_extra_opts:-""}\""
```
tomponline, line 92: Why "extra_opts" rather than making it a dedicated debug option?

tomponline: I'm not sure of the convention with the LXD snap package, but we don't have so far an equivalent way of passing arbitrary flags to the LXD binary, so it may be this was restricted by design. CC @stgraber

tomponline:
Maybe this should just be controlled by the existing `daemon.debug` option?

snapcraft/hooks/configure
mihalicyn: Because we may want to pass other libfuse options, like " -o attr_timeout".

mihalicyn:
> Maybe this should just be controlled by the existing `daemon.debug` option?

I thought about that too, but it's a rare case when you want to debug lxd at the same time as lxcfs. But lxcfs debug mode makes a serious performance impact.

snapcraft/hooks/configure
mihalicyn: > I'm not sure of the convention with the LXD snap package, but we don't have so far an equivalent way of passing arbitrary flags to the LXD binary, so it may be this was restricted by design. CC @stgraber

understand. That's why code review is important :-) I'm newbie with snap, so...

mihalicyn: As I can see we have a very convenient trick for LXD https://github.com/lxc/lxd-pkg-snap/commit/8f340a6ce50cf295f48410a8be2b9ade74c6e651
I think it's worth adding an analogical thing for lxcfs. Then it will be okay just to have snap `lxcfs.debug` option, and if we need something extra, we can just override lxcfs binary with bash script with arbitrary options passed to the lxcfs binary.

tomponline: Yes the sideloading functionality is very useful. Although it does require that the binary is built within an equivalent environment as the snap provides (so shared libraries can link correctly).

stgraber:
To sideload lxcfs, you can just bind-mount an alternative binary over /snap/lxd/current/bin/lxcfs.
To override the startup flags, you can copy, modify and then bind-mount /snap/lxd/current/commands/daemon.start

Having a `lxcfs.debug` config key seems fine to me as that's a useful thing for our users to be able to set.
An arbitrary way to pass in command arguments is probably better done with a temporary override of daemon.start.

snapcraft/hooks/configure
mihalicyn: yep, it will work only if built from ubuntu 20 env. But that's better than nothing! ;-)

mihalicyn:
Ok, so, I'll create another PR for `lxcfs.debug`.

PR #112, opened 2023-01-11
daemon.start: add lxcfs.debug config option

mihalicyn:


stgraber:
Needs to be added to snapcraft.yaml too

mihalicyn:
> Needs to be added to snapcraft.yaml too

ugh, right. Thanks!

snapcraft.yaml
```diff
@@ -43,6 +43,7 @@ description: |-
    - lxcfs.pidfd: Start per-container process tracking [default=false]
    - lxcfs.loadavg: Start tracking per-container load average [default=false]
    - lxcfs.cfs: Consider CPU shares for CPU usage [default=false]
+   - lxcfs.debug: Enable lxcfs debug mode (passes -d option) [default=false]
```
stgraber: I'd probably match the LXD one and say `Increase logging to debug level`.
The user doesn't know or care that it's the `-d` flag to LXCFS and lxcfs is already implied by the option being called `lxcfs.debug` :)

mihalicyn: fixed! :-)

PR #113, opened 2023-01-11
snapcraft: remove strip for liblxc

mihalicyn:
liblxc is small even with debuginfo it's about 400 KiB.

This simplifies coredumps analysis.


mihalicyn:
Btw, we are not stripping liblxc at this point, cause it is placed in `/lib/x86_64-linux-gnu` instead of `/lib`. I'm not sure if it's a bug or not.

PR #114, opened 2023-01-12
daemon.start: use libsegfault for lxcfs

mihalicyn:
This option enables preloading of libSegFault.so from glibc. It allows us to easily catch and print backtrace + register values when crash occurs.

It doesn't prevent from coredump generation if needed.

Stacktrace can be read from:
journalctl -u snap.lxd.daemon.service

Example:
** Segmentation fault
147658]: Register dump:
147658]:  RAX: 0000000000000005   RBX: 00005562952d1820   RCX: 00000000ffffe000
147658]:  RDX: 0000000000000000   RSI: 00007f24fa6b4ec9   RDI: 00007f24ec000c30
147658]:  RBP: 00007f24f9d86b40   R8 : 0000000000000000   R9 : 00007f24fa68c6e8
147658]:  R10: 0000556294dde145   R11: 0000000000000007   R12: 00000000000
...

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -523,6 +523,11 @@ else
         [ "${lxcfs_cfs:-"false"}" = "true" ] && lxcfs_args="${lxcfs_args} --enable-cfs"
         [ "${lxcfs_debug:-"false"}" = "true" ] && lxcfs_args="${lxcfs_args} -d"
 
+        # Preload libSegFault.so to catch crashes
+        export LD_PRELOAD="${LD_PRELOAD:+${LD_PRELOAD}:}/lib/${ARCH}/libSegFault.so"
+        export SEGFAULT_USE_ALTSTACK=1
+        export SEGFAULT_SIGNALS="all"
+
         if [ -n "${lxcfs_args}" ]; then
             # shellcheck disable=SC2086
             lxcfs ${lxcfs_args} "${SNAP_COMMON}/var/lib/lxcfs" -p "${SNAP_COMMON}/lxcfs.pid" &


mihalicyn:
IMHO, it's really useful until we resolve the problem with the `core_pattern` virtualization. It looks fully safe to enable it by default, but let's have it disabled at first. We can ask users who get crashes to enable this if changing `core_pattern` is not possible for some reason or is inconvenient.

stgraber:
Any reason not to simply do this as part of lxcfs.debug? Feels like we're likely to always ask for both.

mihalicyn:
> Any reason not to simply do this as part of lxcfs.debug? Feels like we're likely to always ask for both.

`lxcfs.debug` enables debug mode in libfuse, which produces several `printf` for each fuse request. In a heavy-loaded production environment, this can be dangerous. But `lxcfs.catchsegv` looks safe in all cases.

stgraber:
Ok, then we should just do it by default if it's safe. At least in the edge channel.

mihalicyn:
Agreed. Do I need to change the `default` value in this PR or the default value for the edge channel handled externally?

stgraber:
I wouldn't even put a config key at that point, if we want it on by default, just have the startup script set it up.

Config keys are visible to user in a lot of places, so we're trying to keep them tidy and to what we expect users to consider when configuring the snap. As this one should be safe by default, we should just do it and not offer a config key for it.

mihalicyn:
Ah, understand. Ok, I think we can just enable this by default for now.
In the newer Glibc version they decided to deprecate `libSegFault.so` https://github.com/bminor/glibc/commit/65ccd641bacea33be23d51da737c2de7543d0f5e
It was moved to a separate package https://launchpad.net/ubuntu/+source/glibc-tools

We can consider other options in the future like gnulib sigsegv or https://www.gnu.org/software/libsigsegv/

mihalicyn:
I've updated PR.

PR #117, opened 2023-03-21
edk2-vars-generator: fix flash size for OVMF_4M flavor

monkz:
QemuEfiFlashSize was still at SIZE_2M, while SIZE_4M is already available

### DIFF
--- a/edk2-vars-generator/edk2-vars-generator
+++ b/edk2-vars-generator/edk2-vars-generator
@@ -96,7 +96,7 @@ if __name__ == '__main__':
             'QemuCommand': Qemu.QemuCommand(
                 QemuEfiMachine.OVMF_Q35,
                 variant=QemuEfiVariant.SECBOOT,
-                flash_size=QemuEfiFlashSize.SIZE_2MB,
+                flash_size=QemuEfiFlashSize.SIZE_4MB,
                 code_path=args.code,
                 vars_template_path=args.vars_template,
             ),


stgraber:
No, we can't do that.

monkz:
because of https://bugs.launchpad.net/ubuntu/+source/edk2/+bug/1885662 ?

stgraber:
Changing the firmware to do 4MB NVRAM will break every single existing instance which was created with a firmware using a 2MB NVRAM.

We're aware of the need to rollout 4MB NVRAM to fix issues with modern Windows UEFI revocation lists, but this can't just be a simple change like that. Instead we're going to have to build the firmware twice, once in 2MB and once in 4MB, then make LXD detect the presence of the 4MB variant and prefer it for new VMs which using the 2MB variant for existing ones.

Then a few years from now we may drop support for 2MB firmware with an upcoming LTS (won't be LXD 6.0 but maybe 7.0 can drop 2MB completely).

monkz:
would be exposing this via a "volatile.ovmf_flashsize" be possible? default: 2M
and still put the 4M Version in the craft?

even serving the 4M Version alone (without volatile..), would enable people to tinker with "qemu.raw"

stgraber:
Current plan is to not have it be user visible. We'll just assume that qemu.nvram is 2MB, qemu.4MB.nvram will be 4MB, then LXD detects which one it has for the instance and loads the matching firmware.

A user modifying `security.secureboot` always gets an NVRAM reset so touching that key would effectively convert an existing VM from 2MB to 4MB.

monkz:
@stgraber looks like OVMF_4M is never called https://github.com/lxc/lxd-pkg-snap/blob/54d069cd9479b974ed1050b7222d2478cf5dbda1/snapcraft.yaml#L858

PR #118, opened 2023-03-24
Provide multiple OVMF Builds in varying fd sizes

monkz:
This should enable lxd to switch VMs with secure boot to bigger FD sizes, as it is required for some Windows installations to load more into nvram.


monkz:
If someone could have a look on the aarch64 part - that would be nice. I have problems with cross compiling

stgraber:
Per https://bugs.launchpad.net/ubuntu/+source/edk2/+bug/1885662/comments/9 it suggests that arm64 is fine as it is today.
I don't think we should be needing any change to the python script as the script we're shipping is the exact same used by Debian and Ubuntu today to provide both 2M and 4M builds.

monkz:
I rolled back the changes in edk2-vars-generator and ensured that filenames are similar to the Debian and Ubuntu ovmf package.

monkz:
Currently not ready for merge

PR #119, opened 2023-03-27
Provide multiple OVMF Builds in varying fd sizes

monkz:
This should enable lxd to switch VMs with secure boot to bigger FD sizes, as it is required for some Windows installations to load more into nvram.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -288,55 +288,100 @@ parts:
       git config user.email "noreply@linuxcontainers.org"
       git config user.name "LXD snap builder"
     override-build: |-
-      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
+      OVMF_TARGET_DIRECTORY="${CRAFT_PART_INSTALL}/share/qemu/"
+      CLEAN_AFTER_BUILD=true
+      IFS='' read -r DEFAULT_BUILD_FLAGS << EOF1
+              -DSMM_REQUIRE=TRUE \
+              -DSECURE_BOOT_ENABLE=TRUE \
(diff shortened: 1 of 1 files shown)


monkz:
Sorry #118 auto closed because of git clean up mistake.

This PR is now ready. I tested the resulting snap with a modified lxd version that detects the nvram file and chooses the correct OVMF_CODE.

Corresponding lxd PR is https://github.com/lxc/lxd/pull/11515

stgraber:
Concept looks good. I'd just want to stay away from bashisms as there's no guarantee that snapcraft will run through bash and in general we tend to keep all our shell code to basic POSIX shell.

In this case, the way I'd have done it is by defining functions. In the edk2 part, you'd want a function that's something like `build_edk2 <target> <extra build flags>...` and then call it once as `build_edk2 /path/to/OVMF_CODE.bin -DFD_SIZE_2MB` and once as `build_edk2 /path/to/OVMF_CODE.4MB.bin -DFD_SIZE_4MB` with the 4MB call being conditional on us running on x86_64.

Same deal for the SecureBoot variables, you'd want a function which kinda looks like `build_vars /path/to/OVMF_CODE.bin /path/to/OVMF_VARS.bin`.

Effectively the same result but without needing to use loops or deal with bashisms.

monkz:
Removed the loops  and added the signed-off

stgraber:
I've taken that branch and made a variation on it as part of work I'm doing to enable CSM too.

The resulting will be 3 firmware builds:
 - Regular 2MB
 - Regular 4MB
 - CSM 4MB

I'm now working on the LXD side of it, also re-using some of the work you've done on that side but also adding all the more complex plumbing needed to handle CSM correctly.

monkz:
Thank you.

PR #120, opened 2023-03-29
snapcraft: add libSegFault.so to the prime step

mihalicyn:
My previous fix (https://github.com/lxc/lxd-pkg-snap/pull/116) was insufficient, unfortunately.
Let's add libSegFault.so to the prime step and adjust path to it in LD_PRELOAD env variable.


mihalicyn:
cc @simondeziel

simondeziel (approved):

PR #123, opened 2023-06-23
daemon.start: Fix reading from files under /proc

simondeziel:
dash's `read` implementation only reads 1 byte
from files under /proc, see:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=595063

This is a partial revert of commit 7afc0bc.


simondeziel:
@stgraber this is ready for review, thanks!

PR #127, opened 2023-07-07
github: Tweak SSH integration with Launchpad

simondeziel:
This essentially hardcodes the expected key for git.launchpad.net meaning that it will stop working if it's ever changed. This is however the price to pay to have it's authenticity checked.


tomponline:
@simondeziel please can you open a PR for this against latest-candidate too

PR #130, opened 2023-07-07
editor: only spawn EDITOR from host if unpriv user NS are enabled

simondeziel:


snapcraft/wrappers/remote-viewer
```diff
@@ -45,7 +45,7 @@ if [ "${USERNS}" = "0" ]; then
     echo ""
     echo "Your system has unprivileged user namespaces disabled, as a result, "
     echo "you will need to manually run one a SPICE client with the URI above."
-    sleep infinity
+    exec sleep infinity
```
tomponline, line 48: Why is this change needed?

simondeziel: That's not needed but it caught my eye. The `exec` ensures no `sh` parent sticks around for no purpose because obviously, the `sleep` is terminal (nothing will execute after) and will never return anyway.

monstermunchkin (approved):

PR #131, opened 2023-07-07
snapcraft: update git config user.email

simondeziel:


tomponline:
same again for latest-candidate please

PR #136, opened 2023-08-07
lxd: enable LXD static website to snap

gabrielmougard:
This is a follow up PR linked to https://github.com/canonical/lxd/pull/12098

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1341,6 +1341,8 @@ parts:
       - libacl1-dev
       - libudev-dev
       - pkg-config
+      - shellcheck
+      - python3.10-venv
     build-snaps:
       - go
     stage-packages:
@@ -1393,6 +1395,13 @@ parts:
       go build -o "${CRAFT_PART_INSTALL}/bin/lxd-benchmark" github.com/canonical/lxd/lxd-benchmark
       go build -o "${CRAFT_PART_INSTALL}/bin/lxd-user" github.com/canonical/lxd/lxd-user
 
+      # Build the static website
+      make doc
+
+      # Copy the static website
+      mkdir -p "${CRAFT_PART_INSTALL}/share/lxd-documentation"
+      cp -a doc/html/. "$CRAFT_PART_INSTALL/share/lxd-documentation/"
+
       # Setup bash completion
       mkdir -p ${CRAFT_PART_INSTALL}/etc/bash_completion.d/
       cp scripts/bash/lxd-client ${CRAFT_PART_INSTALL}/etc/bash_completion.d/snap.lxd.lxc
@@ -1420,6 +1429,7 @@ parts:
       - etc/bash_completion.d/snap.lxd.lxc
 
       - share/misc
+      - share/lxd-documentation*
 
       - bin/lxc
       - bin/lxc-to-lxd
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -438,6 +438,10 @@ if [ "${ui_enable:-"false"}" = "true" ]; then
     export LXD_UI="${SNAP_CURRENT}/share/lxd-ui/"
 fi
 
+# Setup documentation
+echo "==> Exposing LXD documentation"
+export LXD_DOCUMENTATION="${SNAP_CURRENT}/share/lxd-documentation"
+
 # Setup shiftfs
 if [ "${shiftfs_enable:-"auto"}" = "auto" ]; then
     echo "==> Disabling shiftfs on this kernel (auto)"


snapcraft.yaml
```diff
@@ -1393,6 +1393,9 @@ parts:
       go build -o "${CRAFT_PART_INSTALL}/bin/lxd-benchmark" github.com/canonical/lxd/lxd-benchmark
       go build -o "${CRAFT_PART_INSTALL}/bin/lxd-user" github.com/canonical/lxd/lxd-user
 
+      # Copy the pre-compressed static website
+      cp -R github.com/canonical/lxd/lxd/gendocs/html "${CRAFT_PART_INSTALL}/bin/"
```
simondeziel: I don't think `/bin` is the right location. The LXD UI is put in `"${CRAFT_PART_INSTALL}/share/lxd-ui/"` for example.
I suspect that Tom probably has an idea of where that should go.

tomponline: When does gendocs directory get populated?

gabrielmougard: there would be a new commit check so gendocs is already populated when you checkout the repo

tomponline: Hrm so that would mean we have to commit the (still pretty large ui docs). Im thinking we should do it as a build step in the snap package too.

gabrielmougard: I think it's actually better to do this yes (what if we decide to have heavy assets like videos in the future... the repo will be heavy)

gabrielmougard:
@tomponline I tested it locally with
```yaml
lxd:
    source: https://github.com/gabrielmougard/lxd-fork
    source-branch: feat/serve-doc-in-browser
```

and it behaves as expected

tomponline (commented):
Thanks for this. 

How much does this increase the resulting snap package size (as opposed to the installation size)?

gabrielmougard:
> Thanks for this.
> 
> How much does this increase the resulting snap package size (as opposed to the installation size)?

Around 74MB

tomponline:
> > Thanks for this.
> > How much does this increase the resulting snap package size (as opposed to the installation size)?
> 
> Around 74MB

Is that the size increase of the .snap file or the directory that gets built?

gabrielmougard:
> > > Thanks for this.
> > > How much does this increase the resulting snap package size (as opposed to the installation size)?
> > 
> > 
> > Around 74MB
> 
> Is that the size increase of the .snap file or the directory that gets built?

Nevermind, this is the folder.. let me check again

tomponline:
> Nevermind, this is the folder.. let me check again

Still that's ludicrously big. Rather than immediately going to compression we should understand why thats so large and whether there is anything we can delete from the built directory that isn't needed when viewing the rendered files.

gabrielmougard:
@tomponline the size of the new `.snap` package is 187M. Where can I check the size of the original one ?

gabrielmougard:
@tomponline and I removed the `.doctrees` folder which is around 30M but useless for us

tomponline:
> 

what is the size of the docs directory after removing the doctrees?

gabrielmougard:
After removing the doctrees and doing `lxd/doc/html $ du -sh .` I get `74M`

tomponline:
@gabrielmougard ok so 74MB is very different from 5MB previously mentioned here:

https://discourse.ubuntu.com/t/expose-the-full-lxd-documentation-website-as-part-of-a-lxd-deployment/37516/11?u=tomp

Whats differing here?

simondeziel:
I thought that using pre-compression was at some point abandoned for 2 reasons:

* client not supporting GZIP would need the server side to deflate prior to sending
* anything already gzip'ed wouldn't benefit from the snap's XZ compression which would have otherwise been better

@gabrielmougard sorry if I missed some discussions where those shortcomings were addressed.

tomponline:
> anything already gzip'ed wouldn't benefit from the snap's XZ compression which would have otherwise been better

Yeah im trending that way myself, just waiting for @gabrielmougard research to come back.

tomponline:
That and removing as much of the duplicated inline css/js/svg from the pages as possible.

gabrielmougard:
@tomponline good news :)
```
(base) gab@gab-desktop:~/go/src/github.com/lxd-pkg-snap$ du -sh lxd_0+git.645ca70_amd64.snap 
184M    lxd_0+git.645ca70_amd64.snap
(base) gab@gab-desktop:~/go/src/github.com/lxd-pkg-snap$ du -sh lxd_0+git.248cd66_amd64.snap 
176M    lxd_0+git.248cd66_amd64.snap
```

I suppose XZ is doing a good job and keep us below the 10MB size increase

tomponline:
Good stuff!

snapcraft.yaml
```diff
@@ -1393,6 +1393,15 @@ parts:
       go build -o "${CRAFT_PART_INSTALL}/bin/lxd-benchmark" github.com/canonical/lxd/lxd-benchmark
       go build -o "${CRAFT_PART_INSTALL}/bin/lxd-user" github.com/canonical/lxd/lxd-user
 
+      # Build the static website
+      apt install -y shellcheck python3.10-venv
```
tomponline: that this is the only occurrence of `apt` in this file should be an alarm to us that we have got this step wrong.

It looks like we need to use `build-packages` instead e.g  https://github.com/canonical/lxd-pkg-snap/pull/136/files#diff-e68b9afd67bb77dec8eda77796d3e0b784c7011203114996383eed6533be0cefL1340

snapcraft.yaml
```diff
@@ -1393,6 +1393,15 @@ parts:
...
+      make doc
+      rm -Rf doc/html/.doctrees
```
tomponline: @ru-fu did you mention that .doctrees will not be in the `html` dir soon?

snapcraft.yaml
```diff
@@ -1393,6 +1393,15 @@ parts:
...
+
+      # Copy the static website
+      mkdir -p "${CRAFT_PART_INSTALL}/share/lxd-documentation"
+      cp -a doc/html/. "$CRAFT_PART_INSTALL/share/lxd-documentation/"
```
tomponline, line 1403: does the lxd-ui part put the `html` dir inside `share/lxd-ui` or does it put the html files directly into `share/lxd-ui`?

snapcraft.yaml
ru-fu: It's the `-d` option I keep talking about ... ;)
I've now updated it in https://github.com/canonical/lxd/pull/12215/commits/2c39e99295555c4f460de2aee3fb74eb78234ab3 (part of https://github.com/canonical/lxd/pull/12215).

tomponline: @ru-fu OK thanks, so there's nothing we need to do here and we can remove the `rm` command and it just wont be in the `html` dir anymore?

ru-fu: Correct.

snapcraft.yaml
gabrielmougard: html files directly inside `share/lxd-ui`

gabrielmougard: Hence why we do the same here

tomponline: OK cool, I like that works, as I was expecting to see a `-r` flag on that.

tomponline (approved):

PR #144, opened 2023-09-07
edk2: disable EFI memory attributes protocol

### LINKED ISSUE #12211: Booting VMs on Arm64 with standard Ubuntu images leads to `Synchronous Exception`
# Required information

 * Distribution: Ubuntu
 * Distribution version: 22.04
 * The output of "lxc info" or if that fails:
   * Kernel version: `6.2.0-31-generic #31~22.04.1-Ubuntu `
   * LXC version: 5.0.3
   * LXD version: 5.17-e5ead86
   * Storage backend in use: dir

# Issue description

Booting a VM on an Arm64 system with the standard Ubuntu images is broken on latest 5.17. I was able to reproduce this on 5.15 and 5.16. First version not affected is 5.14 (from `5.14/candidate` as there is no `5.14/stable` for some reason). VMs are failing with a `Synchronous Exception` right after VM start and when it tries to boot from disk:

```
BdsDxe: loading Boot0001 "UEFI QEMU QEMU HARDDISK " from PciRoot(0x0)/Pci(0x1,0x1)/Pci(0x0,0x0)/Scsi(0x0,0x1)
BdsDxe: starting Boot0001 "UEFI QEMU QEMU HARDDISK " from PciRoot(0x0)/Pci(0x1,0x1)/Pci(0x0,0x0)/Scsi(0x0,0x1)


Synchronous Exception at 0x000000007C318000


Synchronous Exception at 0x000000007C318000
```

Interestingly the issue does not exist when UEFI is trying to do a network boot via MAAS. Also the same issue does not exist on x86.

# Steps to reproduce

 1. Install LXD and initialize `snap install lxd --channel=5.17/stable ; lxd init --auto`
 2. Launch a VM and attach to the console `lxc launch ubuntu:j j0 --vm -c security.secureboot=false`

# Information to attach

 - [ ] Container log (`lxc info NAME --show-log`)
 - [ ] Container configuration (`lxc config show NAME --expanded`)
 - [ ] Main daemon log (at /var/log/lxd/lxd.log or /var/snap/lxd/common/lxd/logs/lxd.log)
 - [ ] Output of the client with --debug
 - [ ] Output of the daemon with --debug (alternatively output of `lxc monitor` while reproducing the issue)

mihalicyn:
Unfortunately, we have to disable EFI memory attributes protocol that was introduced in
https://github.com/tianocore/edk2/commit/1c4dfadb4611ef511816dfdfbdb37d7d100b5a4b starting from edk2-stable202305 as it leads to crash of SecureBoot shim

There is a fix for shim that addresses this issue, but it will take a few month until this fix will be landed to different Linux distros and we can't make our users wait for it.
Fix for shim:
https://github.com/rhboot/shim/commit/c7b305152802c8db688605654f75e1195def9fd6

Fixes https://github.com/canonical/lxd/issues/12211

### DIFF
--- a/patches/edk2-0006-disable-EFI-memory-attributes-protocol.patch
+++ b/patches/edk2-0006-disable-EFI-memory-attributes-protocol.patch
@@ -1,0 +1,30 @@
+From cb5e0080ffd3f522f83b8e9273eac10e132ce7c7 Mon Sep 17 00:00:00 2001
+From: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>
+Date: Thu, 7 Sep 2023 09:07:08 +0200
+Subject: [PATCH] edk2: disable EFI memory attributes protocol
+
+https://github.com/canonical/lxd/issues/12211
+
+Signed-off-by: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>
+---
+ ArmPkg/Drivers/CpuDxe/CpuDxe.c | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/ArmPkg/Drivers/CpuDxe/CpuDxe.c b/ArmPkg/Drivers/CpuDxe/CpuDxe.c
+index d04958e79e..c01d571379 100644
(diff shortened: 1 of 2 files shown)


tomponline (approved):
Loving your work, thanks!

PR #146, opened 2023-09-08
snapcraft: do not build embedded doc for LXD if arch is riscv

gabrielmougard:


tomponline (changes requested):
Please can you refactor this to be an if statement so that we dont accidentally skip future bits added to the lxd build on risc. Also it would be useful in the future if you added a comment to the if statement explaining why we skip docs build on riscv.

snapcraft.yaml
```diff
@@ -1572,6 +1573,13 @@ parts:
           strip -s ${CRAFT_PRIME}/${zfs}/bin/* ${CRAFT_PRIME}/${zfs}/lib/*
       done
 
+      if [ "$(uname -m)" != "riscv64" ]; then
+          # Prime the documentation only if the arch is not riscv64.
+          # Some python dependencies are not available for riscv64 or just require a build from source.
+          # Not worth the effort for now.
+          cp -r $CRAFT_STAGE/share/lxd-documentation $CRAFT_PRIME/share/
```
tomponline: Please use `${CRAFT_PRIME}` like the other places above for consistency.

tomponline: oh and `${CRAFT_STAGE}`.

snapcraft.yaml
```diff
@@ -1403,12 +1403,14 @@ parts:
...
+      if [ "$(uname -m)" != "riscv64" ]; then
+        # Build the static website
+        make doc
 
-      # Copy the static website
-      mkdir -p "${CRAFT_PART_INSTALL}/share/lxd-documentation"
-      cp -a doc/html/. "$CRAFT_PART_INSTALL/share/lxd-documentation/"
+        # Copy the static website
+        mkdir -p "${CRAFT_STAGE}/share/lxd-documentation"
+        cp -a doc/html/. "$CRAFT_STAGE/share/lxd-documentation/"
```
tomponline: Please use consistent format, so here it should be `${CRAFT_STAGE}`

snapcraft.yaml
```diff
@@ -1572,6 +1573,13 @@ parts:
           strip -s ${CRAFT_PRIME}/${zfs}/bin/* ${CRAFT_PRIME}/${zfs}/lib/*
       done
 
+      if [ "$(uname -m)" != "riscv64" ]; then
+          # Prime the documentation only if the arch is not riscv64.
+          # Some python dependencies are not available for riscv64 or just require a build from source.
+          # Not worth the effort for now.
+          cp -r ${CRAFT_STAGE}/share/lxd-documentation ${CRAFT_PRIME}/share/
```
tomponline: Nearly there. :) You should really be using `""` to wrap your variables to prevent potential command injection.
You've done it for the `cp` and `mkdir` calls above, but haven't maintained that detail here.

tomponline (approved):

PR #149, opened 2023-09-18
Content interface for MicroCeph

masnax:
Adds support in LXD for `MicroCeph`s `ceph-conf` slot.

This will add a directory under `/var/snap/lxd/` containing the data in microceph at `/var/snap/microceph/current`. When the slot is connected to the plug, the `connect` hook runs, setting the appropriate `MicroCeph` symlink. On disconnect, it resets the symlink back to `hostfs` or `builtin`.

There's some quirks around auto-connection because it's only supposed to work with the same publisher, and `--devmode` snaps don't have a publisher so I can't actually test that until this is merged into `latest/edge`. For that reason I've split removing the corresponding blocks in `daemon.start` to a different PR.

### DIFF
--- a/snapcraft/hooks/connect-plug-ceph-conf
+++ b/snapcraft/hooks/connect-plug-ceph-conf
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+# Utility functions
+get_bool() {
+    value=$(echo "${1:-}" | tr '[:upper:]' '[:lower:]')
+
+    # See if it's true
+    for yes in "true" "1" "yes" "on"; do
+        if [ "${value}" = "${yes}" ]; then
+            echo "true"
+            return
+        fi
+    done
+
+    # See if it's false
+    for no in "false" "0" "no" "off"; do
+        if [ "${value}" = "${no}" ]; then
+            echo "false"
+            return
+        fi
+    done
+
+    # Invalid value (or not set)
+    return
+}
+
+ceph_builtin=$(get_bool "$(snapctl get ceph.builtin)")
+
+if ! [ "${ceph_builtin:-"false"}" = "true" ]; then
+  ln -snf ${SNAP_DATA}/microceph/conf/ /etc/ceph
+fi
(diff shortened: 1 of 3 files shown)


snapcraft/hooks/connect-plug-ceph-conf
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+ln -snf ${SNAP_DATA}/microceph/conf/ /etc/ceph
```
tomponline: I suppose we shouldn't do this if `ceph.builtin` is enabled?

tomponline: @masnax please can you fix this ASAP so I can merge and you can test a latest/edge build for your testing.

masnax: Fixed now, I had to copy over a helper from `configure` into each hook because I couldn't figure out how to share function and variable definitions between hooks.

masnax: Oh, I just noticed I didn't actually add the commit :)

masnax: Fixed now. @tomponline this should be ready now, thanks.

tomponline (approved):

tomponline:
I'm going to merge this now for testing in latest/edge but if it works I would like to see the get_bool function moved into a common file and sourced where its used.

masnax:
> I'm going to merge this now for testing in latest/edge but if it works I would like to see the get_bool function moved into a common file and sourced where its used.

I had tried this, it seems the scripts actually are run from `/var/snap/lxd/current` so for this to work we would need to write the common functions within the `/var/snap` directory for LXD so we can source them each time a hook runs. 

Technically, the hooks themselves seem to be stored in `/snap/lxd/current/meta` and `/snap/lxd/current/snap`, not sure of the difference between those, but the hook files are written to both. I'm not sure if snaps are normally allowed to access there, it always works with `--devmode`, but since LXD has extra privileges, we may be able to source directly from there and avoid the extra step of writing the common script when the snap runs.

tomponline:
@masnax please can you ask the snap team for advice on this. Thanks

PR #157, opened 2023-09-21
snapcraft/hooks: Add back common file

masnax:
To add to the mess, looks like the common file got dropped in my previous commit. Sorry about that.


masnax:
@tomponline @simondeziel Could one of you merge this?

PR #159, opened 2023-09-21
snapcraft/hooks: Revert split into common file

masnax:
This reverts the sharing of function definitions via sourcing the `common` file, as it seems we can't put non-reserved scripts in the `hooks` directory.

In the meantime I've reverted to just duplicating the functions in each hook. 

As mentioned on mattermost, I can't seem to find a good way to test this without actually using a published snap. The errors we are seeing don't seem to get picked up by `snapcraft lint lxd.broken.snap` or `snap install lxd --stable && snap install ./lxd.broken.snap --dangerous` so it seems trial-and-error is going to be the way forward for now.


simondeziel (approved):
LGTM but on an unrelated note:

```
$ git grep -F verify_int
snapcraft/hooks/common:verify_int () {
```

PR #160, opened 2023-09-22
snapcraft/hooks: Exec interface hooks outside of apparmor confinement

masnax:
`edge` is once again broken, but now only when `microceph` is already installed.

Another error that couldn't be picked up by a local install with snap install --dangerous

The hooks fail with the error:
```
- Run hook connect-plug-ceph-conf of snap "lxd" (run hook "connect-plug-ceph-conf": ln: failed to create symbolic link '/etc/ceph': Permission denied)
```

So this commit adds a line that execs them outside the apparmor confinement to get around the error, similar to the `configure` hook.

### DIFF
--- a/snapcraft/hooks/connect-plug-ceph-conf
+++ b/snapcraft/hooks/connect-plug-ceph-conf
@@ -1,4 +1,10 @@
 #!/bin/sh
+set -eu
+
+# Re-exec outside of apparmor confinement
+if [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current)" != "unconfined" ]; then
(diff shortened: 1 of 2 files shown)


tomponline:
Ah yeah because the whole daemon.start command runs outside of apparmor confinement too:

https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/commands/daemon.start#L6

Out of interest, how did you recreate that error? Is it broken in edge only when microceph is installed, as you say?

tomponline (approved):

masnax:
Reproducer is:
`snap install microceph`
`snap install lxd --edge`

Then you get the error. At least this 100% confirms the auto-connect does indeed work!

PR #163, opened 2023-09-22
MicroCeph content interface

masnax:
There's 2 changes here relative to the previous implementation:

1) There are added entries in `snapcraft.yaml` for the hooks `connect-plug-ceph-conf` and `disconnect-plug-ceph-conf` which add the `lxd-support` and `system-observe` plugs, allowing super-privileged access to call `aa-exec -p unconfined`. 

2) MicroCeph handling is removed from `daemon.start` since we have confirmed that autoconnection to the microceph snap does indeed work.

On point 1, I think there was a bit of user-error in my final attempt last night. I've been installing the snaps in a VM with 

```
snap remove lxd --purge ; snap install lxd ; snap install --dangerous /root/lxd.broken.snap
``` 
but I may have used the wrong snap last night, because rebuilding and trying again today, I can actually see the error:

```
- Run hook connect-plug-ceph-conf of snap "lxd" (run hook "connect-plug-ceph-conf":
-----
cat: /proc/self/attr/current: Permission denied
/snap/lxd/x1/snap/hooks/connect-plug-ceph-conf: 6: exec: aa-exec: Permission denied
-----)
```

Which means this one isn't actually hidden behind publishing like I thought. Can't say the same for the other errors, so I've still left out the `common` file handling until we can test that locally.

I've also confirmed that with the additional lines for `lxd-support` and `system-observe`, I don't get that error when the hooks fire.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -59,6 +59,11 @@ source-code: https://github.com/canonical/lxd
 website: https://ubuntu.com/lxd
 confinement: strict
 
+plugs:
+ ceph-conf:
+   interface: content
+   target: "$SNAP_DATA/microceph"
+
 apps:
   # Main commands
   activate:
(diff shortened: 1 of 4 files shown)


masnax:
Adding you as a reviewer so it's on your radar, but no rush to push this till next week :)

tomponline (approved):

PR #165, opened 2023-09-23
snapcraft: drop `lxd-migrate` (deb -> snap) 

simondeziel:


tomponline (commented):
The last time I brought this up with Stephane he mentioned it would be useful for transitioning from debians lxd into snap lxd. So we should discuss whether we want to support that before removing this.

simondeziel:
Yeah I thought of that but proposed it anyway because the tool specifically checked for Ubuntu only:

```
     // Confirm we're on Ubuntu
     if osID() != "ubuntu" {
             return fmt.Errorf("Data migration is only supported on Ubuntu at this time.")
     }
```

I'll close this as indeed there might be a desire to expand it to cover LXD in Debian or even Incus' deb at some point.

PR #171, opened 2023-09-28
snapcraft/commands: Setup /etc in common file

masnax:
Moves setup of the LXD snap's internal `/etc` into a common file so it can be run both on `daemon.start` and `daemon.activate`


masnax:
I'll undraft this once `snapcraft` is finished locally and I've confirmed it's not going to blow up. It's more likely that if it's going to fail, it would happen in the snap dashboard like last time anyway, but just to be safe.

masnax:
@tomponline This should be ready to test now. 

There's 3 points here that make me a bit nervous about how it'll behave on edge, but I can't verify it locally:
* I've seen env files put under `snapcraft` and be sourced from `${SNAP}/file.env` but I don't know if it'll be safe to call executable scripts from this location without triggering some error
* I'm manually calling the content interface (dis)connection hooks from `setup-etc` so we don't duplicate the logic in them, we'll see if this is allowed.
* This is the biggest one. I have no clear way of testing autoconnection of interfaces locally, so I have no idea if there's going to be a race between `daemon.activate` and the interface hooks if `microceph` is already installed, and therefore autoconnects to LXD.

masnax:
If all goes well here, we can do the same thing for some other paths as well (namely `/run`).

snapcraft/setup-etc
```diff
@@ -0,0 +1,99 @@
...
+# Setup a functional /etc
+echo "==> Preparing a clean copy of /etc"
+
+## Unmount and replace with an empty tmpfs
+if [ -e "/etc/.lxd_generated" ]; then
+    umount -l /etc
+fi
+
+mount -t tmpfs tmpfs /etc -o mode=0755
+touch /etc/.lxd_generated
```
tomponline: I'd like to try and change as little as possible to avoid introducing new unexpected behaviors.

Can we just do the mount of etc and the touch file check in the common part and then leave the rest of this in daemon start. 

Infact daemon.start can resetup etc again from fresh using its existing detection of ceph and ovn.

We only need the content hooks to do something if lxd has already started and etc is mounted.

So we should be able to make the hooks detect if daemon.start has run already using that .lxd_generated file as s hint.

masnax: > We only need the content hooks to do something if lxd has already started and etc is mounted.

masnax: > Can we just do the mount of etc and the touch file check in the common part and then leave the rest of this in daemon start.

Good point, I'll do that, thanks.

> Infact daemon.start can resetup etc again from fresh using its existing detection of ceph and ovn.
>
> We only need the content hooks to do something if lxd has already started and etc is mounted.

What's frustrating about this is we would have to maintain two definitions of what qualifies at MicroCeph being available: whether we can see the state directory, or whether the content interface is connected. If LXD is not running when the connection hooks fire, then once it starts LXD will then overwrite whatever happened in `/etc` as it'll get remade. 

I'd prefer to streamline this by changing `daemon.start` to at least just call the connect & disconnect hooks for the interface depending on what the snap is reporting about the connection status. 

Currently I've set this up using a potentially brittle

```
if snap connections lxd | grep 'ceph-conf' | awk '{print $3}' | grep -q ':ceph-conf' ; then
``` 

But if it goes well then maybe we can do something a bit more robust to where we write to `/var/snap/current/.ceph-conf` or something when the content interface hooks fire, and then `daemon.start` can check for this to determine which hook to fire if it remakes `/etc`.

snapcraft/setup-etc
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+set -eu
+
+# Re-exec outside of apparmor confinement
```
tomponline: We shouldnt need to do this in this script as the callers should have already done it.

masnax: Fixed.

snapcraft/setup-etc
tomponline: So my understanding of this is that there are separate issues here:

1. LXD needs a way to detect if microceph is installed on daemon.start: Currently we use the present of a directory as a test, but it sounds like `snap connections lxd` may also be an approach. I'm not fussed which one we use, if one is more reliable than the other then we should use that one.
2. On daemon.start LXD should maintain settings up its /etc tmpfs mount, as there's a lot going on in that process and its nice to have the behaviour of completely resetting it by running `snap restart lxd` - something that I understand we would lose if we moved it to `daemon.activate`. However we can modify that process to add a file in the /etc mount that the content interface hooks can detect (if needed).
3. The content interface hooks should detect if LXD is running (and thus has potentially missed the detection of the newly installed package). If its not running (can we detect this by the absence of a file in the /etc mount or even detect if /etc isnt mounted?), then we don't need to anything right?
4. If the content interface hooks run and detect LXD is running, then they should update the /etc/ceph symlink if necessary.

masnax: > So my understanding of this is that there are separate issues here:
> 
>     1. LXD needs a way to detect if microceph is installed on daemon.start: Currently we use the present of a directory as a test, but it sounds like `snap connections lxd` may also be an approach. I'm not fussed which one we use, if one is more reliable than the other then we should use that one.
>
>     2. On daemon.start LXD should maintain settings up its /etc tmpfs mount, as there's a lot going on in that process and its nice to have the behaviour of completely resetting it by running `snap restart lxd` - something that I understand we would lose if we moved it to `daemon.activate`. However we can modify that process to add a file in the /etc mount that the content interface hooks can detect (if needed).


The hooks themselves will only ever fire as soon as the connection is established or removed, which should be as soon as the second snap is installed (in which ever order), or whenever it's manually changed with `snap connect` or `snap disconnect`, or if one of the snaps is uninstalled.

`snap connections lxd` reports which interfaces on LXD are connected to which other snaps. So this can be used to detect if the lxd content interface has connected to microceph, which should happen automatically as soon as both services are installed. 

The gripe I have with this is that the output isn't really conducive to scripting, so I've had to pipe it through 2 `grep`s and `awk`. If the output changes, we would have to change this too. The other approach is to have the interface hooks write `connected` or `disconnected` to a file under `/var/snap/current` so `daemon.start` can check this file and then manually exec the hook corresponding to `connect` or `disconnect`.  

>     3. The content interface hooks should detect if LXD is running (and thus has potentially missed the detection of the newly installed package). If its not running (can we detect this by the absence of a file in the /etc mount or even detect if /etc isnt mounted?), then we don't need to anything right?
> 
>     4. If the content interface hooks run and detect LXD is running, then they should update the /etc/ceph symlink if necessary.

The content interface `connect` hook is always going to run before LXD starts, because the autoconnection happens as soon as the second snap has finished installing. So what we need isn't for the hooks to detect if LXD is running, but for `daemon.start` to detect if the interfaces have connected, and then manually fire the correct hook once LXD starts.

tomponline: > The other approach is to have the interface hooks write `connected` or `disconnected` to a file under `/var/snap/current` so `daemon.start` can check this file and then manually exec the hook corresponding to `connect` or `disconnect`.

I'm happy with this if it works OK.

> The content interface `connect` hook is always going to run before LXD starts

Thats not true? The issue we're trying to fix here is when microceph is installed *after* LXD starts.
Installing microceph before LXD starts isn't the issue right (this was working before)?

> So what we need isn't for the hooks to detect if LXD is running, but for `daemon.start` to detect if the interfaces have connected, and then manually fire the correct hook once LXD starts.

Yes we could do that to avoid duplication, but its not 100% necessary so if it is an issue then don't get hung up on it.
My point is that if daemon.start is already correctly detecting if microceph is installed when it starts, then the only issue we are trying to fix is when microceph is installed *after* LXD starts.

masnax: > Thats not true? The issue we're trying to fix here is when microceph is installed _after_ LXD starts. Installing microceph before LXD starts isn't the issue right (this was working before)?

No, that's the opposite. If LXD has started, then `daemon.start` must have run and created the `tmpfs` for us, so MicroCeph shouldn't have any problems setting up its own `/etc/ceph`. The issue happens if we install MicroCeph _before_ LXD starts, which fires the `connect` hook to create a symlink at `/etc/ceph` before we've created the `tmpfs`.

tomponline: If you're able to take a call this would be quicker to resolve, I think we are at crossed purposes.
But its late for you, so dont worry if not able to :)

tomponline (approved):
LGTM!

PR #172, opened 2023-10-02
CI improvements

simondeziel:


.github/workflows/builds.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           TARGET: ${{ github.ref_name }}
         run: |
           set -x
+          git config --global transfer.fsckobjects true
```
tomponline, line 44: What does this do/prevent?

simondeziel: This is a setting that can catch malformed objects and prevent/mitigate certain attacks. It was mentioned as a mitigation to some CVEs (like https://securityaffairs.com/73051/hacking/cve-2018-11235-git-flaw.html). https://resources.infosecinstitute.com/topics/security-awareness/security-best-practices-for-git-users/ recommends setting it (practice 19) so does https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=813157. FYI, setting `transfer.fsckobjects` also provides default values for `fetch.fsckObjects` and `receive.fsckObjects`.

tomponline (approved):

.github/workflows/builds.yml
tomponline: @simondeziel thanks could you also apply this change to the lxd, microcloud and microcloud-pkg-snap repos?

simondeziel: Done:
https://github.com/canonical/lxd/pull/12342
https://github.com/canonical/microcloud/pull/188
https://github.com/canonical/microcloud-pkg-snap/pull/4

PR #174, opened 2023-10-03
qemu: Bump to v8.0.5

mihalicyn:
A little bit later we can update to v8.1.1 I think.


tomponline:
@mihalicyn thanks for this. I do a monthly deps update so would pull this in normally anyway.

PR #176, opened 2023-10-04
Misc improvements

simondeziel:
Some `override-build` sections contain currently unneeded `git config`. Those were left as they could be useful if/when cherry-picks are needed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -327,6 +327,7 @@ parts:
     override-pull: |-
       [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
       set -ex
+      git config --global transfer.fsckobjects true
       git clone https://github.com/tianocore/edk2 . -b edk2-stable202305
 
       git config user.email "noreply@lists.canonical.com"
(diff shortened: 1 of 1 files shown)


tomponline (approved):

tomponline:
@simondeziel this PR has broken the edge build by the looks of it:

https://launchpadlibrarian.net/690296759/buildlog_snap_ubuntu_jammy_s390x_lxd-latest-edge_BUILDING.txt.gz

```
Executing parts lifecycle: pull libmnl
Executing action
Failed to pull source: command ['git', 'clone', '--recursive', '--branch', 'libmnl-1.0.5', 'https://git.netfilter.org/libmnl', '/build/lxd/parts/libmnl/src'] exited with code 128.
Make sure sources are correctly specified.
```

I'll retry them to just check it wasnt a temporary network glitch (we see a lot of them on launchpad)

tomponline:
I think this was actually just a temporary network glitch, apologies for the noise.

simondeziel:
I'm glad it was just a network glitch. The edk2 section is one I had problems with as well. Specifically, I wanted to undo the "manual" `git clone` and relying on snapcraft doing it but it didn't work because we later on do a `git revert` which wouldn't find the commit in question. Are you aware of why we do a `git revert` in that specific case rather than using a patch file like we do for other changes to edk2?

Maybe @mihalicyn would know?

mihalicyn:
>Maybe @mihalicyn would know?

because it is a clean patch revert, in other cases we have a custom patches (not just reverts).

PR #178, opened 2023-10-04
Strip binaries and libraries

simondeziel:
Most binaries and a few libs were already being stripped but many big ones were not:

|Bin/lib|Orig size (M)|Stripped size (M)|
|--------|-----------------|-----------------------|
|qemu-system-x86_64|64|18|
|qemu-img|11|1.9|
|virtiofsd|8.1|2.2|
|sqlite3|6.7|1.5|
|lxc-to-lxd|29|21|
|ovn-sbctl|4.4|3.6|

Unfortunately, the 'biggest offender" (`minio`'s binary) was already stripped so no gain there.

This PR assumes it's OK to be more liberal in what is `strip`'ed because the main binaries (`lxc`, `lxd`, `zfs-*`) were already being stripped. That of course needs to be validate before merging.

Here what it looks when all is packaged up in snaps:

```
root@v1:~# df -ha | grep -wF /snap/lxd
/dev/loop1      112M  112M     0 100% /snap/lxd/24322  # snap from 5.0/stable
/dev/loop4      149M  149M     0 100% /snap/lxd/x1     # snap from this PR
/dev/loop5      182M  182M     0 100% /snap/lxd/25880  # snap from latest/edge
```

So that'd save ~33M compared to `latest/edge` while still being much bigger than `5.0/stable`.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1604,17 +1604,21 @@ parts:
     override-prime: |
       set -x
 
-      # Strip some of the heavy bits
-      strip -s "${CRAFT_PRIME}"/bin/lxc
-      strip -s "${CRAFT_PRIME}"/bin/lxd*
-      strip -s "${CRAFT_PRIME}"/bin/snap*
-      strip -s "${CRAFT_PRIME}"/lib/libdqlite*
-      strip -s "${CRAFT_PRIME}"/lib/libsqlite*
+      # Strip binaries (excluding shell scripts)
+      find "${CRAFT_PRIME}"/bin -type f \
+        -not -path "${CRAFT_PRIME}/bin/ceph" \
+        -not -path "${CRAFT_PRIME}/bin/editor" \
+        -not -path "${CRAFT_PRIME}/bin/lxc-checkconfig" \
+        -not -path "${CRAFT_PRIME}/bin/nvidia-container-cli" \
+        -not -path "${CRAFT_PRIME}/bin/remote-viewer" \
+        -not -path "${CRAFT_PRIME}/bin/sshfs" \
+        -not -path "${CRAFT_PRIME}/bin/xfs_admin" \
+        -exec strip -s {} +
 
-      for zfs in zfs-0.8 zfs-2.0 zfs-2.1 zfs-2.2; do
-          [ ! -d "${CRAFT_PRIME}/${zfs}" ] && continue
-          strip -s "${CRAFT_PRIME}/${zfs}"/bin/* "${CRAFT_PRIME}/${zfs}"/lib/*
-      done
+      # Strip libraries (excluding python3 scripts)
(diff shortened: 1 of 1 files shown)


tomponline (commented):
@simondeziel did this build OK locally and did you test installing it and, launching a VM with QEMU, for a basic test that this hasn't broken anything?

simondeziel:
Yes, this was tested in VM which was then able to launch containers and (nested) VMs:

```
root@v1:~# snap install --dangerous /dev/shm/lxd_0+git.242e454_amd64.snap 
lxd 0+git.242e454 installed
root@v1:~# lxd init --auto
root@v1:~# lxc launch --vm images:alpine/edge av1 -c security.secureboot=false
Command 'lxc' not found, but can be installed with:
snap install lxd            # version 5.18-db8c6f9, or
apt  install lxd-installer  # version 1
See 'snap info lxd' for additional versions.
root@v1:~# snap alias lxd.lxc lxc
Added:
  - lxd.lxc as lxc
root@v1:~# lxc launch --vm images:alpine/edge av1 -c security.secureboot=false
Creating av1
Starting av1                                
root@v1:~# lxc shell av1
Error: LXD VM agent isn't currently running
root@v1:~# lxc shell av1
Error: LXD VM agent isn't currently running
root@v1:~# lxc ls
+------+---------+----------------------+----------------------------------------------+-----------------+-----------+
| NAME |  STATE  |         IPV4         |                     IPV6                     |      TYPE       | SNAPSHOTS |
+------+---------+----------------------+----------------------------------------------+-----------------+-----------+
| av1  | RUNNING | 10.36.207.166 (eth0) | fd42:202c:7e77:a001:216:3eff:fe85:26a (eth0) | VIRTUAL-MACHINE | 0         |
+------+---------+----------------------+----------------------------------------------+-----------------+-----------+
root@v1:~# lxc shell av1
av1:~# 
root@v1:~# 


root@v1:~# lxc launch images:alpine/edge ac1
Creating ac1
Starting ac1                                
root@v1:~# lxc shell ac1
ac1:~# 
```

simondeziel:
I researched it some more and found some old evidences about problems caused by stripping Go binaries (https://bugs.launchpad.net/ubuntu/+source/golang/+bug/1200255 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=717172). Fortunately, Debian (and Ubuntu) reversed this in 2016, https://salsa.debian.org/go-team/compiler/golang/-/commit/4e9a859347c7d8e0ad860a6e4e390d132f178ea1:

>  * Strip the binaries as it has worked for the last five years or so and
>     upstream sees no reason to disable it.

https://honnef.co/articles/stripping-go-elf-binaries-is-safe/ also says it's safe.

I'm leaving this note in case somebody brings this up in the future.

PR #182, opened 2023-10-09
Add support for PowerFlex storage driver

roosterfish:
Allow LXD to communicate with the NVMe/TCP tooling on the host as part of https://github.com/canonical/lxd/pull/12304

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -671,6 +671,12 @@ parts:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container*.so*
 
+  nvme:
+    source: snapcraft/empty
+    plugin: nil
+    stage-packages:
+      - nvme-cli
+
   openvswitch:
     source: https://github.com/openvswitch/ovs
     source-type: git


snapcraft.yaml
```diff
@@ -60,9 +60,15 @@ website: https://ubuntu.com/lxd
...
-   target: "$SNAP_DATA/microceph"
+  ceph-conf:
+    interface: content
+    target: "$SNAP_DATA/microceph"
+  nvme-subsystem:
+    interface: system-files
+    read:
+      - /sys/class/nvme-subsystem
+    write:
+      - /sys/class/nvme-subsystem
```
simondeziel: According to the [doc](https://snapcraft.io/docs/system-files-interface), `write` grants both `read` and `write`.

roosterfish: Thanks, I have now removed `write` completely since only read access to the path is required.

simondeziel: Ah, even better! Thanks

roosterfish: Now the extra plug is completely removed since the access is already provided by the base template. Some more details here: https://chat.charmhub.io/charmhub/pl/hi6qo7gaetd9dkwxe3gph9o8xa

tomponline (approved):

tomponline:
@roosterfish want this merged?

roosterfish:
@tomponline marked it as ready, thanks.

PR #183, opened 2023-10-10
Apparmor unpriv quirk

simondeziel:
Assume `unshare` won't work if `/proc/sys/kernel/apparmor_restrict_unprivileged_userns` or `/proc/sys/kernel/apparmor_restrict_unprivileged_unconfined` exist. Previously we tried to read their value but that fails because those are root-accessible only.


tomponline:
What about if lxd has disabled the feature? Then unpriv userns is still available.

simondeziel:
> What about if lxd has disabled the feature? Then unpriv userns is still available.

The initial PR regressed that case which is unacceptable, thanks for catching. I'll rework that make the default experience nicer without trying to accommodate those undoing sysctl changes done by LXD.

PR #185, opened 2023-10-12
lxd-migrate/golang.org/x/net 0.17.0

simondeziel:


tomponline:
@simondeziel I was thinking we should have the github workflows check that lxd-migrate can build here OK.

simondeziel:
@tomponline not 100% sure that's what you wanted but: https://github.com/canonical/lxd-pkg-snap/pull/186

PR #186, opened 2023-10-12
github: test build lxd-migrate prior to pushing to LP

simondeziel:


.github/workflows/builds.yml
```diff
@@ -56,6 +56,9 @@ jobs:
           rsync -a --exclude .git --delete . ~/lxd-pkg-snap-lp/
           cd ~/lxd-pkg-snap-lp
           lxd-snapcraft -set-version "${ver[0]}" -set-source-commit "${ver[1]}"
+          # Test build lxd-migrate
```
tomponline: Can you add this as a new step between `Install Go` and `Trigger Launchpad snap build` called `Test lxd-migrate build`

tomponline: Oh it should also be a separate job so it runs on PRs rather than just merge pushes.

.github/workflows/builds.yml
```diff
@@ -1,16 +1,40 @@
...
+    name: Test lxd-migrate build
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.18.x
```
tomponline: Can we make this go 1.20 now?

tomponline:
@simondeziel please can you port this to latest-candidate

PR #187, opened 2023-10-20
daemon.start: stop disabling Apparmor restrictions on unpriv userns/unconfined

simondeziel:
Those restrictions are not enabled in 23.10 so LXD no longer need to force disable them. If they are enabled, it means the user opted into it, probably for testing the features, in which case LXD shouldn't undo the user's decision.

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -424,20 +424,6 @@ if [ "$(stat -c '%u' /proc)" = 0 ]; then
             echo 1 > /proc/sys/kernel/unprivileged_userns_clone || true
         fi
     fi
-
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline do you also want a separated PR for `latest-candidate`?

tomponline:
yes please

tomponline:
@alexmurray are you happy for us to merge this and LXD will still work on Mantic?

alexmurray:
Yes this is backwards compatible with the current snapd / kernel behaviour in mantic.

alexmurray:
Oops sorry I thought this was a different PR. Give me a sec to check it m

alexmurray:
Yes this is fine to merge, there is no plan to enable the restriction on mantic currently.

tomponline:
@simondeziel please port to latest-candidate

PR #189, opened 2023-11-06
snapcraft/{hooks,commands}: handle new AppArmor unconfined profile mode

alexmurray:
AppArmor supports a new unconfined profile mode which essentially acts like the unconfined label - when a profile is in this mode the label contains the suffix "(unconfined)" - so treat this the same as the "unconfined" label and don't try and break out of confinement in this case.

This is designed to go hand-in-hand with the proposed changes for snapd in https://github.com/snapcore/snapd/pull/13333

### DIFF
--- a/snapcraft/commands/lxc
+++ b/snapcraft/commands/lxc
@@ -2,8 +2,11 @@
 set -eu
 
 # Re-exec outside of apparmor confinement
-if [ -d /sys/kernel/security/apparmor ] && [ "$(while read -r l; do echo "$l"; done < /proc/self/attr/current)" != "unconfined" ]; then
-    exec /usr/bin/aa-exec -p unconfined -- "$0" "$@"
+if [ -d /sys/kernel/security/apparmor ]; then
+  label="$(cat /proc/self/attr/current)"
+  if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
+    exec aa-exec -p unconfined -- "$0" "$@"
+  fi
 fi
 
 # Check if native and snap installed
--- a/snapcraft/commands/lxc
+++ b/snapcraft/commands/lxc
@@ -3,7 +3,7 @@ set -eu
 
 # Re-exec outside of apparmor confinement
 if [ -d /sys/kernel/security/apparmor ]; then
-  label="$(cat /proc/self/attr/current)"
+  label="$(while read -r l; do echo "$l"; done < /proc/self/attr/current)"
   if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
     exec aa-exec -p unconfined -- "$0" "$@"
   fi
--- a/snapcraft/commands/lxc
+++ b/snapcraft/commands/lxc
@@ -5,7 +5,7 @@ set -eu
 if [ -d /sys/kernel/security/apparmor ]; then
   label="$(while read -r l; do echo "$l"; done < /proc/self/attr/current)"
   if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
-    exec aa-exec -p unconfined -- "$0" "$@"
+    exec /usr/bin/aa-exec -p unconfined -- "$0" "$@"
   fi
 fi
 
--- a/snapcraft/commands/buginfo
+++ b/snapcraft/commands/buginfo
@@ -3,7 +3,7 @@ set -u
 
 # Re-exec outside of apparmor confinement
 if [ -d /sys/kernel/security/apparmor ]; then
-  label="$(cat /proc/self/attr/current)"
+  label="$(cat /proc/self/attr/current 2>/dev/null)"
   if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
     exec aa-exec -p unconfined -- "$0" "$@"
   fi
--- a/snapcraft/commands/buginfo
+++ b/snapcraft/commands/buginfo
@@ -2,8 +2,11 @@
 set -u
 
 # Re-exec outside of apparmor confinement
-if [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current)" != "unconfined" ]; then
+if [ -d /sys/kernel/security/apparmor ]; then
+  label="$(cat /proc/self/attr/current)"
+  if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
     exec aa-exec -p unconfined -- "$0" "$@"
+  fi
 fi
 
 # Check that we're root
--- a/snapcraft/commands/daemon.activate
+++ b/snapcraft/commands/daemon.activate
@@ -2,7 +2,9 @@
 set -eu
 
 # Re-exec outside of apparmor confinement
-if [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current 2>/dev/null)" != "unconfined" ]; then
+if [ -d /sys/kernel/security/apparmor ]; then
+  label="$(cat /proc/self/attr/current)"
+  if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
     if ! aa-exec --help >/dev/null 2>&1; then
         echo "The LXD snap was unable to run aa-exec, this usually indicates a LXD sideload." >&2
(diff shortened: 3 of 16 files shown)


snapcraft/commands/lxc
```diff
@@ -2,8 +2,11 @@
 set -eu
 
 # Re-exec outside of apparmor confinement
-if [ -d /sys/kernel/security/apparmor ] && [ "$(while read -r l; do echo "$l"; done < /proc/self/attr/current)" != "unconfined" ]; then
-    exec /usr/bin/aa-exec -p unconfined -- "$0" "$@"
+if [ -d /sys/kernel/security/apparmor ]; then
+  label="$(cat /proc/self/attr/current)"
+  if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
+    exec aa-exec -p unconfined -- "$0" "$@"
+  fi
```
simondeziel, line 9: The `while read -r` dance and the fully qualified path were used here to minimize the overhead of the `lxc` command that is already slow due to snap confinement. If we could keep that while supporting the new unconfined label it would be nice.

simondeziel (commented):
LGTM except for a small optimization for the `lxc` command. Thanks!

snapcraft/commands/lxc
alexmurray: Can you explain a bit more about why there was a need for the `while` loop here? Was it to read the label without having to spawn a new subprocess?

alexmurray: https://github.com/canonical/lxd-pkg-snap/pull/189/commits/1465c1d764ef67e4efec5fdb41affaa217b5a1f9 gives slightly better performance:

```
[amurray:~] $ time (i=0; while [ $i -lt 1000 ]; do label=$(while read -r l; do echo $l; done < /proc/self/attr/current); i=$((i+1)); done)

real	0m1.435s
user	0m0.762s
sys	0m0.710s
[amurray:~] $ time (i=0; while [ $i -lt 1000 ]; do label=$(cat /proc/self/attr/current); i=$((i+1)); done)

real	0m2.786s
user	0m1.250s
```

But we could do this even faster if the commands used `bash` instead of `sh` for the interpreter, then you could read it as just `label=$(</proc/self/attr/current)`:

```
[amurray:~] $ time (i=0; while [ $i -lt 1000 ]; do label=$(</proc/self/attr/current); i=$((i+1)); done)

real	0m0.010s
user	0m0.000s
sys	0m0.010s
```

simondeziel: Indeed, `dash`'s limitation makes it annoying. Historically we've stuck with a POSIX shell but we should re-assess that decision. That's something I'll look into and deal with in a different PR if we make the switch. Thanks!

snapcraft/commands/lxc
```diff
@@ -2,8 +2,11 @@
 set -eu
 
 # Re-exec outside of apparmor confinement
-if [ -d /sys/kernel/security/apparmor ] && [ "$(while read -r l; do echo "$l"; done < /proc/self/attr/current)" != "unconfined" ]; then
-    exec /usr/bin/aa-exec -p unconfined -- "$0" "$@"
+if [ -d /sys/kernel/security/apparmor ]; then
+  label="$(while read -r l; do echo "$l"; done < /proc/self/attr/current)"
+  if [ "$label" != "unconfined" ] && [ -n "${label##*(unconfined)}" ]; then
+    exec aa-exec -p unconfined -- "$0" "$@"
```
simondeziel:
```suggestion
    exec /usr/bin/aa-exec -p unconfined -- "$0" "$@"
```

This was added to avoid PATH lookup overhead. I don't know how big of a difference it makes but might as well keep it, if you don't mind.

simondeziel (commented):
One last nitpick regarding fully qualified `aa-exec` call.

simondeziel (approved):
Thank you.

simondeziel:
@alexmurray this fails on current Noble image as `/proc/self/attr/current` isn't readable:

```
root@v1:~# snap install --dangerous /dev/shm/lxd_0+git.6d2b48a_amd64.snap 
- Start snap "lxd" (unset) services (systemctl command [start snap.lxd.activate.service] failed with exit status 1: stderr:
Job for snap.lxd.activate.service failed because the control process exited with error code.
See "systemctl status snap.lxd.activate.service" and "journalctl -xeu snap.lxd.activate.service" for details.)
root@v1:~# systemctl status snap.lxd.activate.service
× snap.lxd.activate.service
     Loaded: not-found (Reason: Unit snap.lxd.activate.service not found.)
     Active: failed (Result: exit-code) since Fri 2023-11-10 08:17:29 UTC; 39s ago
   Main PID: 1085 (code=exited, status=1/FAILURE)
        CPU: 28ms

Nov 10 08:17:29 v1 systemd[1]: Starting snap.lxd.activate.service - Service for snap application lxd.activate...
Nov 10 08:17:29 v1 lxd.activate[1104]: cat: /proc/self/attr/current: Permission denied
Nov 10 08:17:29 v1 systemd[1]: snap.lxd.activate.service: Main process exited, code=exited, status=1/FAILURE
Nov 10 08:17:29 v1 systemd[1]: snap.lxd.activate.service: Failed with result 'exit-code'.
Nov 10 08:17:29 v1 systemd[1]: Failed to start snap.lxd.activate.service - Service for snap application lxd.activate.
```

This was tested with:

```
$ lxc launch ubuntu-daily:24.04 v1 --vm
$ lxc exec v1 bash
# pull updates, reboot
# switch to snapd's edge snap
# install lxd snap built from that PR
```

Additional information:

```
root@v1:~# uname -a
Linux v1 6.5.0-9-generic #9-Ubuntu SMP PREEMPT_DYNAMIC Sat Oct  7 01:35:40 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

root@v1:~# snap list snapd
Name   Version                Rev    Tracking     Publisher   Notes
snapd  2.61+git1481.g7a33013  20625  latest/edge  canonical✓  snapd
```

alexmurray:
Hmm that is a bit weird as I don't think I have introduced anything new when reading from /proc/self/attr/current - other than removing the `2>/dev/null` - so has this always failed but now we just see the error message?

alexmurray:
Restored the 'redirect stderr to /dev/null' approach in [1be7337](https://github.com/canonical/lxd-pkg-snap/pull/189/commits/1be7337f4724f2c81f8d0b28964fa5a8f1431d8b)

simondeziel:
Sorry, I was not testing the hand built snap properly. I had not installed the official snap prior to using the snap from this PR which meant it couldn't properly install (missing `ack` or `connections`?). Anyway, with that solved, I see no error at install and will be doing proper testing sometimes next week.

simondeziel:
@tomponline please merge when you can.

tomponline:
@simondeziel please can you port this to latest-candidate

snapcraft/commands/lxc
simondeziel: @tomponline we briefly discussed and agreed to switching to `bash` for the snap but I just ran those quick tests and think we should stick to `dash` for now.

For some reason, I cannot reproduce Alex's finding with `bash` efficiently reading from `procfs` in one go.

```
# reading from procfs
$ time /bin/sh -c 'for _ in $(seq 10000); do label=$(while read -r l; do echo $l; done < /proc/self/attr/current); done'

real	0m2.465s
user	0m1.662s
sys	0m0.969s

$ time /bin/bash -c 'for _ in $(seq 10000); do label=$(</proc/self/attr/current); done'

real	0m4.370s
user	0m3.186s
sys	0m1.505s
```

Either way, I think most of the "overhead" comes from the shell startup cost:

```
# spawning a shell
$ time (for _ in $(seq 10000); do /bin/sh -c 'true'; done)

real	0m11.102s
user	0m8.060s
sys	0m3.902s


$ time (for _ in $(seq 10000); do /bin/bash -c 'true'; done)

real	0m15.712s
user	0m10.653s
sys	0m5.951s
```

Combining both spawning the shell and reading from `procfs` only widens the gap:

```
# spawning a shell and reading from procfs
$ time (for _ in $(seq 10000); do /bin/sh -c 'label=$(while read -r l; do echo $l; done < /proc/self/attr/current)'; done)

real	0m13.735s
user	0m10.532s
sys	0m4.411s

$ time (for _ in $(seq 10000); do /bin/bash -c 'label=$(</proc/self/attr/current)'; done)

real	0m20.139s
user	0m14.123s
sys	0m7.376s
```

PR #191, opened 2023-11-07
lxd-migrate: remove support for upstart

simondeziel:
AFAIK, upstart was used by CentOS before 7 and Ubuntu before 15.10. According to https://en.wikipedia.org/wiki/Upstart_(software) only ChromeOS and ChromiumOS are still using upstart but those are not something distrobuilder support building.

Same was done a while ago in distrobuilder (https://github.com/lxc/distrobuilder/pull/734).


tomponline:
Please can this be applied to latest-candidate branch as a separate pr

PR #193, opened 2023-11-09
snapcraft: start to ship edk2 debug build

mihalicyn:
Related to https://github.com/canonical/lxd/pull/12522

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -371,13 +371,15 @@ parts:
           shift
           TARGET_VARS="$1"
           shift
+          TARGET_BUILD_TYPE="$1"
+          shift
 
           set -ex
           (
           cat << EOF
               . ./edksetup.sh
               make -C BaseTools ARCH="${ARCH}"
-              build -a "${ARCH}" -t GCC5 -b RELEASE -p "${PKG}" \
+              build -a "${ARCH}" -t GCC5 -b "${TARGET_BUILD_TYPE}" -p "${PKG}" \
                 -DSMM_REQUIRE=FALSE \
                 -DSECURE_BOOT_ENABLE=TRUE \
                 -DNETWORK_IP4_ENABLE=TRUE \
@@ -390,8 +392,8 @@ parts:
       EOF
           ) | bash -e
 
-          cp Build/*/*/FV/${FV_CODE}.fd "${TARGET_CODE}"
-          cp Build/*/*/FV/${FV_VARS}.fd "${TARGET_VARS}"
+          cp Build/*/${TARGET_BUILD_TYPE}*/FV/${FV_CODE}.fd "${TARGET_CODE}"
+          cp Build/*/${TARGET_BUILD_TYPE}*/FV/${FV_VARS}.fd "${TARGET_VARS}"
 
           if [ "$(uname -m)" = "aarch64" ]; then
               truncate -s 64m "${TARGET_CODE}"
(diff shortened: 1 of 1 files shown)


mihalicyn:
This is a "universal" version of debug firmware. So if we need some specific build parameters or something else then I will have to build edk2 on my local machine and sign it with my GPG key (or at least publish a hash) and we can ask user to do a hacky machinery like:
```
SNAP_REV=$(snap list lxd | awk '{print $3}' | tail -n 1)
sudo nsenter -m -t $(pidof lxd) mount --bind /path/to/new/firmware/OVMF_CODE.4MB.fd /snap/lxd/${SNAP_REV}/share/qemu/OVMF_CODE.4MB.fd
```
to replace the existing firmware inside the LXD snap in runtime.

snapcraft.yaml
```diff
@@ -422,9 +426,25 @@ parts:
...
             "${CRAFT_PART_INSTALL}/share/qemu/OVMF_VARS.4MB.CSM.fd" \
+            RELEASE \
             -DFD_SIZE_4MB \
             -DCSM_ENABLE=TRUE
       fi
+
+      # Debug firmware (4MB, debug mode)
+      # Set DEBUG_VERBOSE flag
+      if [ "$(uname -m)" = "aarch64" ]; then
+          sed -i "s#DEBUG_PRINT_ERROR_LEVEL = 0x8000004F#DEBUG_PRINT_ERROR_LEVEL = 0x8040004F#g" ArmVirtPkg/ArmVirt.dsc.inc
```
simondeziel, line 428:
```suggestion
          sed -i "s#DEBUG_PRINT_ERROR_LEVEL = 0x8000004F#DEBUG_PRINT_ERROR_LEVEL = 0x8040004F#g" "${PKG}.inc"
```

Since we already have a `PKG` variable, it feels safer to make use of it to avoid anyone changing the variable and forgetting about the hardcoded reference.

snapcraft.yaml
```diff
@@ -422,9 +426,25 @@ parts:
...
+      fi
+      if [ "$(uname -m)" = "x86_64" ]; then
+          sed -i "s#PcdDebugPrintErrorLevel|0x8000004F#PcdDebugPrintErrorLevel|0x8040004F#g" OvmfPkg/OvmfPkgX64.dsc
```
simondeziel:
```suggestion
          sed -i "s#PcdDebugPrintErrorLevel|0x8000004F#PcdDebugPrintErrorLevel|0x8040004F#g" "${PKG}"
```

simondeziel (approved):
Just a tiny nitpick that you can integrate or ignore, I don't feel strongly about it :)

snapcraft.yaml
mihalicyn: Fixed, thanks!

snapcraft.yaml
mihalicyn: It doesn't work, because names are different `ArmVirtPkg/ArmVirtQemu.dsc` and `ArmVirtPkg/ArmVirt.dsc`. There is an additional `Qemu` word.

snapcraft.yaml
```diff
@@ -422,9 +426,25 @@ parts:
           build_edk2 \
             "${CRAFT_PART_INSTALL}/share/qemu/OVMF_CODE.4MB.CSM.fd" \
             "${CRAFT_PART_INSTALL}/share/qemu/OVMF_VARS.4MB.CSM.fd" \
+            RELEASE \
             -DFD_SIZE_4MB \
             -DCSM_ENABLE=TRUE
       fi
+
+      # Debug firmware (4MB, debug mode)
```
tomponline, line 424: Can we add some reference in the comments here to the new config option in lxd that will use thus firmware so its clear the two are related when doing a cross repo code search.

mihalicyn: Agree, have added ;-)

PR #194, opened 2023-11-15
edk2: add custom patch for bootorder support in CSM mode

### LINKED ISSUE #11908: Can't programmatically boot into CSM legacy disk
# Required information

```
  kernel_version: 5.15.0-73-generic
  lxc_features:
    cgroup2: "true"
    core_scheduling: "true"
    devpts_fd: "true"
    idmapped_mounts_v2: "true"
    mount_injection_file: "true"
    network_gateway_device_route: "true"
    network_ipvlan: "true"
    network_l2proxy: "true"
    network_phys_macvlan_mtu: "true"
    network_veth_router: "true"
    pidfd: "true"
    seccomp_allow_deny_syntax: "true"
    seccomp_notify: "true"
    seccomp_proxy_send_notify_fd: "true"
  os_name: Ubuntu
  os_version: "22.04"
  project: default
  server: lxd
  server_clustered: false
  server_event_mode: full-mesh
  server_name: cloudbase-lxd
  server_pid: 130597
  server_version: "5.15"
  storage: zfs
  storage_version: 2.1.5-1ubuntu6~22.04.1
  storage_supported_drivers:
  - name: dir
    version: "1"
    remote: false
  - name: lvm
    version: 2.03.11(2) (2021-01-08) / 1.02.175 (2021-01-08) / 4.45.0
    remote: false
  - name: zfs
    version: 2.1.5-1ubuntu6~22.04.1
    remote: false
  - name: btrfs
    version: 5.16.2
    remote: false
  - name: ceph
    version: 17.2.5
    remote: true
  - name: cephfs
    version: 17.2.5
    remote: true
  - name: cephobject
    version: 17.2.5
    remote: true
```

# Issue description

I successfully booted into a legacy disk, but I need to always configure UEFI settings upon VM creation via console. Here's the VM configuration:
```
architecture: x86_64
config:
  limits.cpu: "6"
  limits.memory: 4096MiB
  security.csm: "true"
  security.secureboot: "false"
  user.user-data: |
    #cloud-config
    users:
      - name: cloudbase
        ssh-authorized-keys:
          - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCh84ov/r/xjOd58tDxx4CGt/WC8gR1Twmb4oP8jLuuFBHbLLHZL5xyVJFZvl+6Ctx29g4lhtWMm43K/m6xqmTk6t2zHRUcskEPBaVdahTvMZ1khb8FWna83FzAWPWlfw74IDx9eOauxrRGawv4NqoPHrEeoQXs53qslep12KRT2Nh4KPACEzcEPOezzNzH6vj9FSpO3cXh6C1BIzfHL2cOLKJ/P4CSFbloaHbZSYrTZvNZmVE6ZVnPr+rpgGmSm+PPHuPQkXvO/w3NOK3buFn46HKEY8m34TXxIIXtgWTnOqS9eZM077ZqoAt7JX/q94z/k6JvwNR9aUt3J8Wn3SmR tmp@migration
        sudo: ['ALL=(ALL) NOPASSWD:ALL']
        groups: sudo
        shell: /bin/bash
  volatile.cloud-init.instance-id: 0f61eac0-5c93-4c9d-8010-741c7dbed366
  volatile.eth0.host_name: tap5fec66b2
  volatile.eth0.hwaddr: 00:16:3e:50:4f:72
  volatile.last_state.power: RUNNING
  volatile.last_state.ready: "false"
  volatile.uuid: 82a7d248-a2de-409b-af45-344572873f6f
  volatile.uuid.generation: 82a7d248-a2de-409b-af45-344572873f6f
  volatile.vsock_id: "145"
devices:
  /dev/sdb:
    boot.priority: "1"
    pool: default
    source: a28253e9-81b0-4cca-ae73-2b784833be9b
    type: disk
  eth0:
    boot.priority: "0"
    name: eth0
    network: lxdbr0
    type: nic
  root:
    boot.priority: "2"
    path: /
    pool: default
    size: 11GiB
    type: disk
ephemeral: false
profiles: []
stateful: false
description: ""
```

The `root` disk of the VM contains the data of a bootable legacy disk, and /dev/sdb is a data disk. When I start this VM, it tries to boot the UEFI disks (as configured), and then using the NICs' PXE/HTTP. Since I have CSM enabled, the legacy boot options are also added, but unfortunately those are by default put at the bottom of the boot priority list.

We'd need to programmatically boot into the legacy disk . Since CSM is enabled in the VM configuration and disk devices set as top priority, it would be nice if the legacy disks were also set as priority next to the UEFI disk entries. 

In my case booting without touching the VM's console, PXE would take over and the VM would never boot into the legacy disk.

mihalicyn:
```
From ("edk2: force CSM boot mode for bootorder") description: 
========
If "-fw_cfg name=opt/com.canonical.lxd/force_csm,string=yes" is specified in the Qemu commandline then treat bootorder settings from Qemu as settings for CSM boot. Right now bootorder from Qemu is always considered as UEFI and CSM boot is always last in the priority list. Even worser, you can't really make CSM to be the first in the bootorder, because in the SetBootOrderFromQemu() function we always fixup bootorder in accordance with Qemu provided and setting from NVRAM are always got reset.

Fortunately, we don't need to patch SeaBIOS, because it already handles bootorder from Qemu properly. So, we just need to pass execution from UEFI to SeaBIOS and from this point everything will work just fine.

This is a very hacky solution but we have no choice here. Because the alternative here is to extend Qemu fwcfg "bootorder" interface to make it aware of CSM/UEFI boot modes and then patch UEFI and SeaBIOS to handle all of that. This can be an ready-to-upstream solution, but I'm not sure that it can be interesting for edk2 folks, because all of this CSM stuff considered as a legacy thing.

This thing is enabled only when CSM_ENABLE=TRUE AND "-fw_cfg name=opt/com.canonical.lxd/force_csm,string=yes" specified in the Qemu cmdline.
========
```

Fixes https://github.com/canonical/lxd/issues/11908
Related to https://github.com/canonical/lxd/pull/12564

### DIFF
--- a/patches/edk2-0007-force-CSM-boot-mode-for-bootorder.patch
+++ b/patches/edk2-0007-force-CSM-boot-mode-for-bootorder.patch
@@ -1,0 +1,115 @@
+From 14b3103f2069194416633447b940c01d31b6682b Mon Sep 17 00:00:00 2001
+From: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>
+Date: Thu, 16 Nov 2023 12:20:18 +0100
+Subject: [PATCH] edk2: force CSM boot mode for bootorder
+
+If "-fw_cfg name=opt/com.canonical.lxd/force_csm,string=yes" is specified
+in the Qemu commandline then treat bootorder settings from Qemu
+as settings for CSM boot. Right now bootorder from Qemu is always considered
+as UEFI and CSM boot is always last in the priority list. Even worser,
+you can't really make CSM to be the first in the bootorder, because in the
+SetBootOrderFromQemu() function we always fixup bootorder in accordance with
+Qemu provided and setting from NVRAM are always got reset.
+
+Fortunately, we don't need to patch SeaBIOS, because it already handles
+bootorder from Qemu properly. So, we just need to pass execution from
+UEFI to SeaBIOS and from this point everything will work just fine.
+
+This is a very hacky solution but we have no choice here. Because
+the alternative here is to extend Qemu fwcfg "bootorder" interface
+to make it aware of CSM/UEFI boot modes and then patch UEFI and SeaBIOS
+to handle all of that. This can be an ready-to-upstream solution,
+but I'm not sure that it can be interesting for edk2 folks, because
+all of this CSM stuff considered as a legacy thing.
+
+This thing is enabled only when CSM_ENABLE=TRUE AND
+"-fw_cfg name=opt/com.canonical.lxd/force_csm,string=yes"
+specified in the Qemu cmdline.
+
+Signed-off-by: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>
+---
+ .../QemuBootOrderLib/QemuBootOrderLib.c       | 48 +++++++++++++++++++
+ .../QemuBootOrderLib/QemuBootOrderLib.inf     |  1 +
+ 2 files changed, 49 insertions(+)
+
+diff --git a/OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.c b/OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.c
+index 2fe6ab30c0..b0efab9b00 100644
+--- a/OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.c
++++ b/OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.c
+@@ -2190,6 +2190,9 @@ SetBootOrderFromQemu (
+   EFI_BOOT_MANAGER_LOAD_OPTION  *BootOptions;
+   UINTN                         BootOptionCount;
+ 
++  FIRMWARE_CONFIG_ITEM  ForceCSMFwCfgItem;
++  UINTN                 ForceCSMFwCfgSize;
++
+   Status = QemuFwCfgFindFile ("bootorder", &FwCfgItem, &FwCfgSize);
+   if (Status != RETURN_SUCCESS) {
+     return Status;
+@@ -2245,6 +2248,51 @@ SetBootOrderFromQemu (
+     goto ErrorFreeBootOptions;
+   }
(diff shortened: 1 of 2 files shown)


mihalicyn:
Let's wait until:
https://github.com/canonical/lxd/pull/12522
https://github.com/canonical/lxd/pull/12515
will be merged

patches/edk2-0007-force-CSM-boot-mode-for-bootorder.patch
```diff
@@ -0,0 +1,115 @@
...
++
++    DEBUG ((DEBUG_VERBOSE, "%a: ForceCSMFwCfg enabled\n", __func__));
++
++    /* make all CSM boot options first in the bootorder list */
++    for (Idx = 0; Idx < ActiveCount; ++Idx) {
++      /* ActiveOption[Idx].BootOption is a CSM option? */
++      if ((DevicePathType (ActiveOption[Idx].BootOption->FilePath) == BBS_DEVICE_PATH) &&
++          (DevicePathSubType (ActiveOption[Idx].BootOption->FilePath) == BBS_BBS_DP))
++      {
++        DEBUG ((DEBUG_VERBOSE, "%a: Found a legacy option: %s\n", __func__, ActiveOption[Idx].BootOption->Description));
```
simondeziel:
```suggestion
+        DEBUG ((DEBUG_VERBOSE, "%a: Found a legacy boot option: %s\n", __func__, ActiveOption[Idx].BootOption->Description));

```

simondeziel (approved):
LGTM for the extent I could review. There are a few minor English mistakes in the patch description which is probably fine as this won't be upstream'ed.

patches/edk2-0007-force-CSM-boot-mode-for-bootorder.patch
mihalicyn: Fixed, thanks!

tomponline (approved):
LGTM thanks!

PR #214, opened 2023-11-28
snapcraft: prime fusermount3 binary

simondeziel:
With the move from `fuse` to `fuse3`, the package content changed:

from `fuse`:

```
/bin/fusermount (binary)
```

to `fuse3`:

```
/bin/fusermount -> fusermount (compat symlink)
/bin/fusermount3 (binary)
```

However, the snapcraft.yaml wasn't updated accordingly meaning that we only shipped the (broken) symlink:

```
$ ll /snap/lxd/current/bin/fusermount*
lrwxrwxrwx 1 root root 11 Mar 23  2022 /snap/lxd/current/bin/fusermount -> fusermount3
```

Note: this also affects the 5.0 branch (I'll send another PR after this one gets merged).

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1348,6 +1348,7 @@ parts:
       usr/lib/: lib/
     prime:
       - bin/fusermount
+      - bin/fusermount3
       - lib/*/libfuse3.so.*
 
       - lib/*/libSegFault.so


simondeziel:
@mihalicyn it seems the problem was introduced by commit 7d6f7407e7 in Oct 2022, I'm surprised nobody noticed/complained for that long. I've only found `fusermount` being used in `daemon.start` and `daemon.stop` and always with failure ignored. Since it's related to `lxcfs`, I'd like to hear your thought on this, thanks!

While you are at it, would you mind checking if the `libfuse.so` bits in (https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/commands/daemon.start#L490 and https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/commands/daemon.start#L499) are still right considering this:

```
$ ll /snap/lxd/current/lib/x86_64-linux-gnu/*fuse*
lrwxrwxrwx 1 root root     18 Mar 23  2022 /snap/lxd/current/lib/x86_64-linux-gnu/libfuse3.so.3 -> libfuse3.so.3.10.5
-rw-r--r-- 1 root root 252088 Nov 28 07:12 /snap/lxd/current/lib/x86_64-linux-gnu/libfuse3.so.3.10.5
```

That doesn't look right to me but you are the SME.

simondeziel:
Or maybe the `libfuse.so` thingy is only meant for migration between `fuse` and `fuse3`? If yes, how long should we carry this?

tomponline (approved):

mihalicyn:
> @mihalicyn it seems the problem was introduced by commit [7d6f740](https://github.com/canonical/lxd-pkg-snap/commit/7d6f7407e7fdc464c38dca192f99bb0d865225c4) in Oct 2022, I'm surprised nobody noticed/complained for that long. I've only found `fusermount` being used in `daemon.start` and `daemon.stop` and always with failure ignored. Since it's related to `lxcfs`, I'd like to hear your thought on this, thanks!

Good catch! It should affect `systemctl stop snap.lxd.daemon` I guess and prevent the LXCFS from being stopped when LXD daemon is stopped.


> That doesn't look right to me but you are the SME.

I agree with you. I think there should be something like `sed -n "/libfuse3\?\.so/ s/.*libfuse3\?\.so\.\([^ ]\+\) .*/\1/p"`

mihalicyn:
> Or maybe the `libfuse.so` thingy is only meant for migration between `fuse` and `fuse3`? If yes, how long should we carry this?

This thing is about handling libfuse minor version upgrades. For example, when 2.7 becomes 2.8 we need to restart the whole LXCFS process to start using a new libfuse version. But this piece of code was not designed (and changed appropriately) to handle libfuse3. That's awesome that you have noticed that. ;)

simondeziel:
> > Or maybe the `libfuse.so` thingy is only meant for migration between `fuse` and `fuse3`? If yes, how long should we carry this?
> 
> This thing is about handling libfuse minor version upgrades. For example, when 2.7 becomes 2.8 we need to restart the whole LXCFS process to start using a new libfuse version. But this piece of code was not designed (and changed appropriately) to handle libfuse3. That's awesome that you have noticed that. ;)

Good that answers my next question about the need to keep this even after the transition to fuse3, thanks! Maybe we could make the sed work with both fuse and fuse3 to make it easier.

mihalicyn:
> Maybe we could make the sed work with both fuse and fuse3 to make it easier.

yes, I guess that this `sed -n "/libfuse3\?\.so/ s/.*libfuse3\?\.so\.\([^ ]\+\) .*/\1/p"` should work for both libfuse version 2 and 3 as I have added a `\?` after a `3`.

PR #219, opened 2023-12-01
lxd-ui: Bump to 0.5 (latest-candidate)

edlerd:
Pulling in a bunch of fixes (see https://github.com/canonical/lxd-ui/releases) for the upcoming 5.20 release


simondeziel:
It seems that the PR title was fixed to include the target branch and then when CI re-ran, the `TITLE` variable still had the old PR title?

PR #237, opened 2023-12-06
Minimize snap staging and restricted fallback editors

simondeziel:
Stop priming binaries that are already available in the base/core snap.  In the case of `vim`, this required to rework how the fallback editor is provided with a `vimrc`.

Not relying on primed binaries means those will be provided by the base snap (`core20` or `core22` ATM) which hopefully will benefit from regular maintenance.

In the case of `vim` and for the `5.0/stable` channel, this would have saved us from using a known vulnerable version:

```
$ grep vim-tiny /snap/core20/current/usr/share/snappy/dpkg.list 
ii  vim-tiny                       2:8.1.2269-1ubuntu5.15       amd64        Vi IMproved - enhanced vi editor - compact version

$ grep '^- vim-tiny' /snap/lxd/current/snap/manifest.yaml 
- vim-tiny=2:8.1.2269-1ubuntu5.11
```

In the above, `core20` ships `... .15` while the `5.0/stable` one has `... .11`.

While in there, I added a restricted mode to our 2 fallback editors. This is a weak security guard rail but has no visible downside from a user's perspective.

This was tested locally to be working.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1477,7 +1477,6 @@ parts:
       - bin/rsync
       - bin/setfacl
       - bin/sgdisk
-      - bin/unsquashfs
       - bin/xdelta3
 
       - lib/*/libidn.so.*
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1024,18 +1024,6 @@ parts:
       - bin/tar2sqfs
       - lib/libsquashfs.so*
 
-  vim:
-    plugin: nil
-    stage-packages:
-      - vim-common
-      - vim-tiny
-    organize:
-      usr/bin/: bin/
-      usr/share/vim/vim*/debian.vim: etc/vimrc
-    prime:
-      - bin/vim.tiny
-      - etc/vimrc
-
   virtiofsd:
     source: https://gitlab.com/virtio-fs/virtiofsd
     source-type: git
@@ -1585,7 +1573,6 @@ parts:
       - sqlite
       - squashfs-tools-ng
       - swtpm
-      - vim
       - virtiofsd
       - xfs
       - xz
--- a/snapcraft/wrappers/editor
+++ b/snapcraft/wrappers/editor
@@ -38,7 +38,8 @@ fi
 # Try running the editor through the host.
 if [ -n "${EDIT_CMD}" ] && [ "${USERNS}" = 1 ]; then
     exec 9< /tmp/
-    EDIT_PATH_HOST="$(echo "${EDIT_PATH}" | sed "s#/tmp/#/proc/self/fd/9/#g")"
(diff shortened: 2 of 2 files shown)


snapcraft.yaml
```diff
@@ -1477,7 +1477,6 @@ parts:
       - bin/rsync
       - bin/setfacl
       - bin/sgdisk
-      - bin/unsquashfs
```
tomponline, line 1480: Is this still getting built somewhere, where did it come from previously?

snapcraft/wrappers/editor
```diff
@@ -42,21 +42,29 @@ if [ -n "${EDIT_CMD}" ] && [ "${USERNS}" = 1 ]; then
     find_and_spawn "${EDIT_CMD}" "${EDIT_PATH_HOST}"
 fi
 
+EDIT_EXTRA=""
```
tomponline: Please could you update the commit message and comments in the code to explain what this is doing and why, I'm afraid I cant follow the reason for this change.

snapcraft.yaml
simondeziel: It is in the core20/core22 snaps, same as `sed` for example.

simondeziel:
```
$ ll /snap/core2?/current/usr/bin/unsquashfs 
-rwxr-xr-x 1 root root 114808 Oct 11  2021 /snap/core20/current/usr/bin/unsquashfs*
-rwxr-xr-x 1 root root 135064 Mar 25  2022 /snap/core22/current/usr/bin/unsquashfs*
```

tomponline: Sorry I meant where does the staged version of bin/unsquashfs come from that is being uncommented here?

simondeziel: The staged package is `squashfs-tools` in the `lxd` part. The binary/package is also in the core bases:

```
$ grep squash /snap/core2?/current/usr/share/snappy/dpkg.list 
/snap/core20/current/usr/share/snappy/dpkg.list:ii  squashfs-tools                 1:4.4-1ubuntu0.3             amd64        Tool to create and append to squashfs filesystems
/snap/core22/current/usr/share/snappy/dpkg.list:ii  squashfs-tools                 1:4.5-3build1                           amd64        Tool to create and append to squashfs filesystems
```

I hope I got your question right this time around.

snapcraft/wrappers/editor
simondeziel: Indeed, I renamed the variable to EDIT_IGNORE_RC which is more descriptive and added a comment. This matches the explanation in the commit, let me know what you think.

tomponline (approved):

snapcraft.yaml
tomponline: Do we need the staged package anymore?

PR #244, opened 2023-12-11
LXD 5.20 sync with latest-edge (latest-candidate)

tomponline:
Will wait until LXD 5.20 is tagged and updated in snapcraft.yaml before merging this.
As the changes depends on code changes in LXD 5.20.


lxd-migrate/go.mod
```diff
@@ -2,40 +2,39 @@ module github.com/canonical/lxd-pkg-snap/lxd-migrate
 
 go 1.18
```
simondeziel: I guess that's something we should think about bumping to `1.20`.

simondeziel (dismissed):

lxd-migrate/go.mod
tomponline: Yes please

simondeziel (dismissed):

simondeziel (approved):

PR #262, opened 2024-01-09
Dynamically lookup the base snap name

simondeziel:
Rather than hardcode the core base name, look it up where/when needed. This was tested locally:

`daemon.start`:

```
$ sudo ls -l /var/snap/lxd/common/mntns/etc/ | grep core
lrwxrwxrwx 1 root root   37 Jan  9 12:48 alternatives -> /snap/core22/current/etc/alternatives
lrwxrwxrwx 1 root root   33 Jan  9 12:48 apparmor -> /snap/core22/current/etc/apparmor
lrwxrwxrwx 1 root root   35 Jan  9 12:48 apparmor.d -> /snap/core22/current/etc/apparmor.d
```

`editor`:

```
$ EDITOR=foo lxc config edit &
$ grep -a 'core[0-9]*' "/proc/$(pgrep vim.tiny)/environ"
LESSOPEN=| /usr/bin/lesspipe %s
HISTTIMEFORMAT=%F %T 
LANGUAGE=en_CA:en
LC_TIME=C.UTF-8
SNAP_COMMON=/var/snap/lxd/common
SNAP_INSTANCE_KEY=USER=sdeziel
XDG_SESSION_TYPE=x11
LXD_CONF=/home/sdeziel/snap/lxd/common/config
LD_LIBRARY_PATH=/var/lib/snapd/lib/gl:/var/lib/snapd/lib/gl32:/var/lib/snapd/void:/snap/lxd/x1/lib:/snap/lxd/x1/lib/x86_64-linux-gnu
SHLVL=0
SNAP_UID=1000
TEMPDIR=/tmp
HOME=/home/sdeziel/snap/lxd/x1
SNAP_LIBRARY_PATH=/var/lib/snapd/lib/gl:/var/lib/snapd/lib/gl32:/var/lib/snapd/void
DESKTOP_SESSION=ubuntu
SNAP_USER_DATA=/home/sdeziel/snap/lxd/x1
...
VIMINIT=source /snap/core22/current/etc/vim/vimrc
...
EDITOR=/snap/lxd/x1/bin/editor foo
```

`kmod`:

```
# nsenter -m -t "$(lxc info | awk '/server_pid:/ {print $2}')" sh -x /usr/sbin/lsmod | grep ^+
+ basename /usr/sbin/lsmod
+ NAME=lsmod
+ sed -n /^name:/ s/^name:\s*\(core[0-9]\{2\}\)/\1/p /meta/snap.yaml
+ SNAP_BASE=core22
+ id -u
[...]

PR #263, opened 2024-01-09
snapcraft/commands/daemon.start: always remount /bin/kmod

simondeziel:
Upon update, the `daemon.start` would keep any existing mountbind of `/bin/kmod` rather than remount a fresh one. This lead to such situation:

```
root@sdeziel-lemur:~# nsenter -m -t "$(lxc info | awk '/server_pid:/ {print $2}')"
root@sdeziel-lemur:/# mount | grep kmod
/var/lib/snapd/snaps/lxd_26615.snap (deleted) on /usr/bin/kmod type squashfs (ro,relatime,errors=continue,threads=single
```

Always remounting that script should avoid holding onto old snap data.

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -136,7 +136,8 @@ fi
 
 # Fix lsmod/modprobe
 echo "==> Setting up kmod wrapper"
-mountpoint -q /bin/kmod || mount -o ro,bind "${SNAP}/wrappers/kmod" "/bin/kmod"
+mountpoint -q /bin/kmod && umount -l /bin/kmod
+mount -o ro,bind "${SNAP}/wrappers/kmod" "/bin/kmod"
 
 # Setup /boot
 if [ -e /var/lib/snapd/hostfs/boot ]; then


tomponline (commented):
Does LXD also need to umount this on LXD stop?

tomponline:
@simondeziel I had a quick search in lxd-pkg-snap for /bin/kmod and noticed there are some hardcoded references to "core22", given we are planning to change to core24 soon, I suspect this may become an issue. Please could you see if there is a way to make that dynamic somehow to refer to the correct core?

tomponline:
@simondeziel ah looks like you had the same though as me here https://github.com/canonical/lxd-pkg-snap/pull/262 thanks!

simondeziel:
> Does LXD also need to umount this on LXD stop?

`daemon.stop` does surprisingly little `umount`'ing, should we also `umount` other bits? @mihalicyn I'd like your input on this, please.

tomponline:
> > Does LXD also need to umount this on LXD stop?
> 
> `daemon.stop` does surprisingly little `umount`'ing, should we also `umount` other bits? @mihalicyn I'd like your input on this, please.

If daemon.stop is called during `systemctl reload snap.lxd.daemon` then we probably should unmount anything as it could be in use by running instances.

simondeziel:
While `daemon.stop` has code meant to handle reloads:

https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/commands/daemon.stop#L32-L42:
```
echo "=> Stop reason is: ${reason}"
...
# Handle reloads and crashes
if [ "${reason}" = "reload" ] || [ "${reason}" = "crashed" ]; then
    exit 0
fi
```

It is not invoked by either `systemctl reload snap.lxd.daemon` or the heavier wait `snap restart lxd --reload` as I'm not seeing the `=> Stop reason is:` message in the logs.

This is probably fine because reloads have a dedicated script (`daemon.reload`, https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft.yaml#L81) which essentially just signals the existing LXD that a reload is needed an then kill it.

tomponline:
Well thats odd that `snap restart lxd` doesnt invoke it, id have thought it would invoke stop and then start hooks (as that does stop instances).

There are clearly differences though as there is this open snapd bug https://bugs.launchpad.net/snapd/+bug/2028141

simondeziel:
> Well thats odd that `snap restart lxd` doesnt invoke it, id have thought it would invoke stop and then start hooks (as that does stop instances).

Did you miss the `--reload` in `snap restart lxd --reload`?

tomponline:
> > Well thats odd that `snap restart lxd` doesnt invoke it, id have thought it would invoke stop and then start hooks (as that does stop instances).
> 
> Did you miss the `--reload` in `snap restart lxd --reload`?

i did :)

tomponline (approved):

PR #266, opened 2024-01-09
snapcraft: ensure logrotate.conf isn't group writable

simondeziel:
Upon startup, LXD might complain during the `==> Rotating logs` part:

```
$ journalctl --no-hostname -b0 -u snap.lxd.daemon --grep logrotate
Jan 09 11:30:33 lxd.daemon[642189]: Potentially dangerous mode on /snap/lxd/x1/etc/logrotate.conf: 0664
Jan 09 11:30:33 lxd.daemon[642189]: error: Ignoring /snap/lxd/x1/etc/logrotate.conf because it is writable by group or others.
```

For some reasons (unknown to me), this only affected snaps I build locally, not those from the snapstore. Here's a debug build with `ls -l` surrounding the `chmod g-w` call:

```
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.592 :: + broken_symlinks=
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.592 :: + '[' -n '' ']'
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.593 :: + ls -l /root/prime/etc/logrotate.conf
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.600 :: -rw-rw-r-- 4 root root 145 Sep 22 21:11 /root/prime/etc/logrotate.conf
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.600 :: + chmod g-w /root/prime/etc/logrotate.conf
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.601 :: + ls -l /root/prime/etc/logrotate.conf
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.602 :: -rw-r--r-- 4 root root 145 Sep 22 21:11 /root/prime/etc/logrotate.conf
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.603 :: + exit 0
2024-01-09 16:07:33.880 :: 2024-01-09 16:07:03.729 patch_elf: not enabled for part 'strip'
```

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1644,6 +1644,8 @@ parts:
           exit 1
       fi
 
+      # Fixup logrotate.conf permissions to not be group writable
+      chmod g-w "${CRAFT_PRIME}/etc/logrotate.conf"
       exit 0
 
   wrappers:


tomponline:
Thanks @simondeziel 

My question would be, why is logrotate.conf even in the snap package? Is it even used by the system logrotate service, or does snap use it for its own log rotation?

simondeziel:
The `daemon.start` script (https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/commands/daemon.start#L340-L342) calls `logrotate`:
```
# Rotate logs
echo "==> Rotating logs"
logrotate -f "${SNAP}/etc/logrotate.conf" -s "/etc/logrotate.status" || true
```
And the `logrotate.conf` file itself (https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/etc/logrotate.conf) manages the rotation of `/var/snap/lxd/common/lxd/logs/lxd.log`.

As for the difference between a local build and one from the builders, I think it boils down to the `umask` used during the `git clone` as my local copy has the same perms (`0664`) on `logrotate.conf` as what ended up in the snap.

tomponline:
Very good, thanks

tomponline (approved):

PR #267, opened 2024-01-16
Use Ubuntu's QEMU source

tomponline:
This updates LXD's snap package to use QEMU from Ubuntu's source: https://git.launchpad.net/ubuntu/+source/qemu

We are initially pinning to 8.1.3+ds-1ubuntu2 which is currently the version in Ubuntu Noble: 

https://packages.ubuntu.com/noble/qemu-system-x86

It uses a tagged release and applies Ubuntu patches to the source before building it.

The ROM files are used from pre-built packages in the current base snap (core22), which will be switched to core24 ones when that is available.

Tested on:

- [x] amd64
- [x] arm64
- [x] s390x
- [x] armhf (build only as no qemu support)

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -811,9 +811,9 @@ parts:
       - libusb
       - spice-protocol
       - spice-server
-    source: https://gitlab.com/qemu-project/qemu
+    source: https://git.launchpad.net/ubuntu/+source/qemu
     source-depth: 1
-    source-tag: v8.1.3
+    source-tag: import/1%8.1.3+ds-1ubuntu2
     source-type: git
     plugin: autotools
     autotools-configure-parameters:
@@ -851,6 +851,7 @@ parts:
       - --enable-virtfs
       - --firmwarepath=/snap/lxd/current/share/qemu/
       - --localstatedir=/var/
+      - --disable-install-blobs # Ubuntu sources don't have the ROM blobs in them.
     build-packages:
       - bison
       - bzip2
@@ -858,21 +859,26 @@ parts:
       - pkg-config
       - libaio-dev
       - libcap-ng-dev
+      - libfdt-dev
       - libglib2.0-dev
       - libnuma-dev
       - libpixman-1-dev
       - librbd-dev
       - libseccomp-dev
       - libusbredirhost-dev
+      - quilt
     stage-packages:
       - genisoimage
+      - ipxe-qemu # This is needed due to --disable-install-blobs.
       - libatomic1
+      - libfdt1
       - libmagic1
       - libnuma1
       - libpixman-1-0
       - libusbredirhost1
       - libusbredirparser1
-      - seabios
+      - seabios # This is needed due to --disable-install-blobs.
+      - qemu-system-data # This is needed due to --disable-install-blobs.
     override-prime: |-
       [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
       craftctl default
@@ -885,16 +891,21 @@ parts:
       # Mangle the configure a bit
       QEMUARCH="$(uname -m)"
       [ "${QEMUARCH}" = "ppc64le" ] && QEMUARCH="ppc64"
+
+      # Apply patches from Ubuntu sources.
+      QUILT_PATCHES=debian/patches quilt push -a
+
       sed -i "s/^unset target_list$/target_list=\"${QEMUARCH}-softmmu\"/" configure
       sed -i 's#libseccomp_minver=".*#libseccomp_minver="0.0"#g' configure
+
+      # Extract efi-virtio.rom from ipxe-qemu.
+      # This doesn't work in the organize section below.
+      mkdir -p "${CRAFT_PART_INSTALL}"/share/qemu
+      mv "${CRAFT_PART_INSTALL}"/usr/lib/ipxe/qemu/efi-virtio.rom "${CRAFT_PART_INSTALL}"/share/qemu/
 
       set +ex
       craftctl default
 
-      set -ex
-      # we don't want to take this file from the qemu tree, but instead from the seabios package
-      rm "${CRAFT_PART_INSTALL}/usr/local/share/qemu/bios-256k.bin"
-      set +ex
     organize:
       usr/bin/: bin/
       usr/lib/: lib/
@@ -902,7 +913,12 @@ parts:
       usr/local/lib/: lib/
       usr/local/libexec/: bin/
       usr/local/share/: share/
+      usr/share/qemu/kvmvapic.bin: share/qemu/
+      usr/share/qemu/s390-ccw.img: share/qemu/
+      usr/share/qemu/s390-netboot.img: share/qemu/
+      usr/share/qemu/slof.bin: share/qemu/
       usr/share/seabios/bios-256k.bin: share/qemu/
+      usr/share/seabios/vgabios-*: share/qemu/
     prime:
       - bin/genisoimage*
       - bin/mkisofs*
@@ -914,13 +930,22 @@ parts:
       - lib/*/libnuma*so*
       - lib/*/libpixman*so*
       - lib/*/libusbredir*so*
-      - share/qemu/keymaps*
-      - share/qemu/efi-virtio.rom*
-      - share/qemu/kvmvapic.bin*
-      - share/qemu/s390-*.img*
-      - share/qemu/slof.bin*
-      - share/qemu/vgabios-*.bin*
-      - share/qemu/bios-256k.bin*
+      - lib/*/libfdt*.so*
+      - share/qemu/keymaps/*
+      - share/qemu/bios-256k.bin
+      - share/qemu/efi-virtio.rom
+      - share/qemu/kvmvapic.bin
+      - share/qemu/s390-ccw.img
+      - share/qemu/s390-netboot.img
+      - share/qemu/slof.bin
+      - share/qemu/vgabios-ati.bin
+      - share/qemu/vgabios-bochs-display.bin
+      - share/qemu/vgabios-cirrus.bin
+      - share/qemu/vgabios-qxl.bin
+      - share/qemu/vgabios-ramfb.bin
+      - share/qemu/vgabios-stdvga.bin
+      - share/qemu/vgabios-virtio.bin
+      - share/qemu/vgabios-vmware.bin
 
   qemu-ovmf-secureboot:
     after:


snapcraft.yaml
```diff
@@ -898,6 +901,10 @@ parts:
       - libpixman-1-0
       - libusbredirhost1
       - libusbredirparser1
+      - libfdt-dev
```
simondeziel: It's surprising to see a `-dev` in the staged packages list. Would you mind adding a comment explaining it's use there, please.

simondeziel: Could we get away with just `libfdt1`? It ships the real `.so` AFAICT:

```
# apt-get install -y libfdt1 libfdt-dev

# ll /usr/lib/x86_64-linux-gnu/libfdt.so
lrwxrwxrwx 1 root root 11 Jan  6  2022 /usr/lib/x86_64-linux-gnu/libfdt.so -> libfdt.so.1
# ll /usr/lib/x86_64-linux-gnu/libfdt.so.1 
lrwxrwxrwx 1 root root 15 Jan  6  2022 /usr/lib/x86_64-linux-gnu/libfdt.so.1 -> libfdt-1.6.1.so

# dpkg -S libfdt.so
libfdt1:amd64: /usr/lib/x86_64-linux-gnu/libfdt.so.1
libfdt-dev: /usr/lib/x86_64-linux-gnu/libfdt.so
# dpkg -S libfdt-1.6.1.so
libfdt1:amd64: /usr/lib/x86_64-linux-gnu/libfdt-1.6.1.so
```

tomponline: Yep good point will update.

snapcraft.yaml
```diff
@@ -878,26 +878,33 @@ parts:
...
       - genisoimage
+      - ipxe-qemu # This is needed due to --disable-install-blobs.
       - libatomic1
       - libmagic1
       - libnuma1
       - libpixman-1-0
       - libusbredirhost1
       - libusbredirparser1
+      - libfdt1
+      - qemu-system-data # # This is needed due to --disable-install-blobs.
```
simondeziel: `s/ # #/ #/`

snapcraft.yaml
```diff
@@ -878,26 +878,33 @@ parts:
...
+      - seabios # This is needed due to --disable-install-blobs.
```
simondeziel, line 880: I just checked and this will get us `1.15.0` while we were using `1.16.2` in `latest/edge`.

snapcraft.yaml
```diff
@@ -878,26 +878,33 @@ parts:
...
+      - libfdt-dev
       - libglib2.0-dev
       - libnuma-dev
       - libpixman-1-dev
       - librbd-dev
       - libusbredirhost-dev
+      - quilt
     stage-packages:
       - genisoimage
+      - ipxe-qemu # This is needed due to --disable-install-blobs.
```
simondeziel, line 872: This package has the same version in 22.04 and 24.04 (`1.21.1+git-20220113.fbbdc3926-0ubuntu1`) which points to this upstream commit https://github.com/ipxe/ipxe/commit/fbbdc3926 from Jan 2022. Upstream is pretty active so maybe we'll be missing on some recent bug fixes when moving from the git submodule to the Ubuntu package.

The change in this Ubuntu specific (not from Debian) package was from Christian Ehrhardt so if we need fresher versions, we could always go to him.

snapcraft.yaml
```diff
@@ -909,18 +916,33 @@ parts:
...
       QEMUARCH="$(uname -m)"
       [ "${QEMUARCH}" = "ppc64le" ] && QEMUARCH="ppc64"
+
+      # Apply patches from Ubuntu sources.
+      QUILT_PATCHES=debian/patches quilt push -a
+
       sed -i "s/^unset target_list$/target_list=\"${QEMUARCH}-softmmu\"/" configure
       sed -i 's#libseccomp_minver=".*#libseccomp_minver="0.0"#g' configure
 
+      # Extract efi-virtio.rom from ipxe-qemu.
```
simondeziel, line 901: There is also `pxe-virtio.rom` from that same package. Dunno if that's something we might be interested in eventually (not in that PR).

snapcraft.yaml
```diff
@@ -932,12 +954,21 @@ parts:
...
+      - share/qemu/s390-netboot.img
+      - share/qemu/slof.bin
+      - share/qemu/vgabios-ati.bin
+      - share/qemu/vgabios-bochs-display.bin
+      - share/qemu/vgabios-cirrus.bin
+      - share/qemu/vgabios-qxl.bin
+      - share/qemu/vgabios-ramfb.bin
+      - share/qemu/vgabios-stdvga.bin
+      - share/qemu/vgabios-virtio.bin
+      - share/qemu/vgabios-vmware.bin
```
simondeziel, line 948: Could we drop the `ati`, `cirrus` and `vmware` ones? For those firmwares to be used, one would need to override the QEMU config and replace the virtio GPU by one of those emulated devices. I can't think of a situation where that would be useful and many of the QEMU CVEs were due to those emulated devices.

The idea would be to limit the user's choice to the saner alternatives GPUs.

simondeziel: I mentioned it here but I think it would belong to a follow-up PR if you agree on the principles of course.

snapcraft.yaml
mihalicyn: That's a good point. Probably we can do something like:
```
git submodule update --init
make -C roms bios
```
(see https://github.com/qemu/qemu/blob/d0f4aa7d50d485b1fb5ec8ab6f934e5df852ab07/roms/Makefile#L55)

This should build SeaBIOS from the submodule https://github.com/qemu/qemu/blob/master/roms/seabios

Resulting file should be in `pc-bios/bios-256k.bin`.

snapcraft.yaml
mihalicyn: The same as above:
```
git submodule update --init
make -C roms pxerom
```

resulting file should be `pc-bios/pxe-virtio.rom`

P.S. Have not tested that.

simondeziel: @mihalicyn ATM, we only care about `efi-virtio.rom` but I'm assuming that's a moot point.

mihalicyn: my bad, sure. Then we need `make -C roms efirom` and `pc-bios/efi-virtio.rom` file.

snapcraft.yaml
tomponline: Yes I made the files match what is currently in the snap in this PR

mihalicyn:
LGTM.

These questions about specific versions of SeaBIOS/pxe/etc is a matter of discussion. I don't know if we want to use just Ubuntu-shipped packages everywhere, or we want to be consistent with the Qemu code base in terms of firmware versions to be choosen.

mihalicyn (dismissed):

snapcraft.yaml
tomponline: We will likely get a version boost when we move to core24 and will then benefit from Ubuntu support of these packages.

tomponline: But yes aware it will go backwards (that's why I made a point of it in the PR description).

tomponline:
> LGTM.
> 
> These questions about specific versions of SeaBIOS/pxe/etc is a matter of discussion. I don't know if we want to use just Ubuntu-shipped packages everywhere, or we want to be consistent with the Qemu code base in terms of firmware versions to be choosen.

Yeah. I'm not sure yet. 

On the one hand the reason we are exploring switching to the ubuntu qemu source is to benefit from ubuntu's maintenance of the qemu packages, whilst still being able to keep reasonably current with upstream and not be stuck on an LTS core.

But WRT to the firmware blobs, they don't necessarily need to be kept in lock step with the qemu version, and if we start checking out the sub-modules (if this is even possible from the ubuntu qemu source) and building those we are then not using the the modules that would be shipped with an ubuntu qemu because they are built in other packages.

Perhaps we can also pull down the source of the other qemu packages and build those.

But there is a certain attraction of building the firmwares that were ratified with the upstream qemu too.

But then we would lose any patches that ubuntu were making to those firmwares.

snapcraft.yaml
tomponline: good spot thanks

tomponline:
I'll ask the Ubuntu server team what they would suggest here.

snapcraft.yaml
tomponline: @simondeziel if we dont need these files we could land a PR now that removes them from the current build and i can then update my PR to reflect that, want to take that one on?

simondeziel: @tomponline sure: https://github.com/canonical/lxd-pkg-snap/pull/284

snapcraft.yaml
```diff
@@ -915,13 +931,22 @@ parts:
...
       - lib/*/libusbredir*so*
+      - lib/*/libfdt*.so*
       - share/qemu/keymaps*
-      - share/qemu/efi-virtio.rom*
-      - share/qemu/kvmvapic.bin*
-      - share/qemu/s390-*.img*
-      - share/qemu/slof.bin*
-      - share/qemu/vgabios-*.bin*
       - share/qemu/bios-256k.bin*
+      - share/qemu/efi-virtio*.rom
```
simondeziel: There is only one `efi-virtio` ROM file so I'd use:
```suggestion
      - share/qemu/efi-virtio.rom
```

snapcraft.yaml
```diff
@@ -915,13 +931,22 @@ parts:
...
       - lib/*/libpixman*so*
       - lib/*/libusbredir*so*
+      - lib/*/libfdt*.so*
       - share/qemu/keymaps*
-      - share/qemu/efi-virtio.rom*
-      - share/qemu/kvmvapic.bin*
-      - share/qemu/s390-*.img*
-      - share/qemu/slof.bin*
-      - share/qemu/vgabios-*.bin*
       - share/qemu/bios-256k.bin*
```
simondeziel: There is only one `bios-256k.bin` file in the seabios package so I'd drop the trailing `*`.

snapcraft.yaml
```diff
@@ -915,13 +931,22 @@ parts:
       - lib/*/libnuma*so*
       - lib/*/libpixman*so*
       - lib/*/libusbredir*so*
+      - lib/*/libfdt*.so*
       - share/qemu/keymaps*
```
simondeziel: This is a directory, so I'd be explicit about it, see https://packages.ubuntu.com/jammy/all/qemu-system-data/filelist
```suggestion
      - share/qemu/keymaps/*
```

simondeziel (dismissed):
LGTM, with few nitpicks.

snapcraft.yaml
tomponline: this is to make them option on the non-qemu archs.

snapcraft.yaml
tomponline: this is to make them option on the non-qemu archs.

tomponline: I think i could further improve the build to not install these packages either on those archs.

snapcraft.yaml
simondeziel: Ah, I thought that for that you needed a `-` prefix like `-bin/vgimportclone` we have elsewhere.

tomponline: maybe thats another way! does that mean optional?

simondeziel: Yes, `-bin/foo` means it will tentatively copy it but won't complain if it cannot.

simondeziel: I couldn't find any official reference though but tests confirmed that `-` makes the copy optional.

mihalicyn (dismissed):

snapcraft.yaml
simondeziel: There's an annoying catch with snapcraft and using selectors (like `on selector1[,selector2]`). The selectors are `AND` instead of `OR`'ed meaning that if you want to make something only for a few arches, you need to duplicate the list like this:

```
build-packages:
  - common-pkg1
  - ...
  - on amd64:
      - virt-enabled-arch-pkg1
      - virt-enabled-arch-pkg2
      - ...
  - on arm64:
      - virt-enabled-arch-pkg1
      - virt-enabled-arch-pkg2
      - ...
  - on s390x:
      - virt-enabled-arch-pkg1
      - virt-enabled-arch-pkg2
      - ...
```

This is explained in https://snapcraft.io/docs/snapcraft-advanced-grammar. That's also the reason we had the `edk2`  part using this:

```
     build-packages:
       - g++
       - on amd64:
         - acpica-tools
         - uuid-dev
       - on arm64:
         - acpica-tools
         - uuid-dev
```
Instead of this:
```
     build-packages:
       - g++
       - on amd64,arm64:
         - acpica-tools
         - uuid-dev
```

So in the end, selectively installing packages for build/stage isn't really convenient :/

tomponline: yeah and it didnt work with the prime section either when i tried it.

tomponline: ill go with the optional indicator instead for prime files

tomponline (commented):
I think the change @simondeziel landed to not prime qemu on unsupported archs means we dont need the optional selectors anymore.

mihalicyn (approved):

PR #271, opened 2024-01-19
snapcraft: remove bios-256k from organize step

mihalicyn:
Remove
      usr/share/seabios/bios-256k.bin: share/qemu/bios-256k.bin
from the "organize" step as it's already in a right place. We will need this once we step on the Ubuntu QEMU builds.


mihalicyn:
Tested with local `snapcraft` build.

simondeziel (approved):

tomponline (approved):

PR #273, opened 2024-01-22
snapcraft: make bios-256k.bin optional at prime step

mihalicyn:
Unfortunately, we have to use this hack to make armhf/riscv builds to succeed, as on these two architectures we don't support QEMU at all. But prime step is still being executed which makes a whole snap build to fail [1], [2].

Also, let's always use bios-256k.bin SeaBIOS from the seabios package instead of from the QEMU tree.

[1] https://launchpadlibrarian.net/710544674/buildlog_snap_ubuntu_jammy_armhf_lxd-latest-edge_BUILDING.txt.gz
[2] https://launchpadlibrarian.net/710258725/buildlog_snap_ubuntu_jammy_riscv64_lxd-latest-edge_BUILDING.txt.gz

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -875,6 +875,11 @@ parts:
 
       set +ex
       craftctl default
+
+      set -ex
+      # we don't want to take this file from the qemu tree, but instead from the seabios package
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks!

PR #277, opened 2024-01-24
snapcraft.yaml: enable unconfined mode in lxd-support interface

alexmurray:
The associated PR for snapd is still open in https://github.com/snapcore/snapd/pull/13514 but opening this one against the lxd snap so the two can be reviewed in tandem.

This should not be merged until at least the snapd one has been AND the correct assumes syntax has been decided upon.

--

The old `lxd-support` interface allowed lxd to set itself to be unconfined - ie to not have any apparmor profile associated with it - but when the userns restrictions are enabled it would then still be blocked from using userns - so we then instead added the unconfined mode which allows a profile to essentially be the equivalent of the unconfined profile (ie no profile) - but then to also add in extra permissions (like the use of userns).

ie. the only way to use userns when the restrictions are enabled is to have an apparmor profile which grants the userns permission in apparmor - AND for the case of lxd which wants to be unconfined we then use the new unconfined profile mode as well

Signed-off-by: Alex Murray <alex.murray@canonical.com>

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,7 +1,7 @@
 name: lxd
 base: core22
 assumes:
-  - snapd2.39
+  - snapd2.62
 version: git
 grade: devel
 summary: LXD - container and VM manager
@@ -71,6 +71,9 @@ plugs:
  ovn-chassis:
    interface: content
    target: "$SNAP_DATA/microovn/chassis"
+ lxd-support-with-unconfined-mode:
+   interface: lxd-support
+   enable-unconfined-mode: true
 
 apps:
   # Main commands
@@ -78,7 +81,7 @@ apps:
     command: commands/daemon.activate
     daemon: oneshot
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
 
   daemon:
@@ -91,7 +94,7 @@ apps:
     slots:
       - lxd
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - network-bind
       - system-observe
     sockets:
@@ -105,7 +108,7 @@ apps:
     restart-condition: on-failure
     daemon: simple
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - network-bind
       - system-observe
     sockets:
@@ -117,60 +120,60 @@ apps:
     command: commands/lxc
     completer: etc/bash_completion.d/snap.lxd.lxc
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
 
   lxd:
     command: commands/lxd
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
 
   # Sub-commands
   buginfo:
     command: commands/buginfo
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   check-kernel:
     command: commands/lxd-check-kernel
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
 
 hooks:
   connect-plug-ceph-conf:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   disconnect-plug-ceph-conf:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   connect-plug-ovn-certificates:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   disconnect-plug-ovn-certificates:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   connect-plug-ovn-chassis:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   disconnect-plug-ovn-chassis:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
   configure:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - network
       - system-observe
   remove:
     plugs:
-      - lxd-support
+      - lxd-support-with-unconfined-mode
       - system-observe
 
 parts:


tomponline:
> assumes syntax

What is this @alexmurray ?

snapcraft.yaml
```diff
@@ -1,7 +1,7 @@
 name: lxd
 base: core22
 assumes:
-  - snapd2.39
+  - snapd2.61
```
simondeziel: The doc isn't really clear on this https://snapcraft.io/docs/snapcraft-yaml-reference:

> A list of features that must be supported by the core in order for this snap to install. For example, to make the snap only installable on certain recent version of snapd (like 2.38) you can specify: - snapd2.38. See [Snapcraft top-level metadata](https://snapcraft.io/docs/snapcraft-top-level-metadata#heading--assumes) for other potential values.

Does that mean if someone just `apt-get install snapd` and got 2.58, then tries to `snap install lxd` that will force snapd to self-refresh and then get at least 2.61?

alexmurray: I am not sure - I think it may prevent it from being installed / snapd may refuse to install it or upgrade to it etc if the feature is not already present (ie if snapd has not already been upgraded) - @pedronis can you provide any clarification on this?

alexmurray: Also I think this would end up being `assumes: [ snapd2.62 ]` instead of 2.61 but will need some clarification from the snapd team on when they feature in the associated snapd PR is likely to land in a released version of snapd.

alexmurray: I just pushed an extra commit to set the assumes to `snapd2.62` as confirmed by @pedronis in https://github.com/snapcore/snapd/pull/13514#issuecomment-1996707279

alexmurray:
Rebased this onto current latest-edge and squashed the two commits into one.

alexmurray:
Hi folks, the snapd support for this feature landed in the 2.62 release of snapd - any chance you could merge this into the lxd snap? cc @tomponline @simondeziel

tomponline:
Hi @alexmurray 

Before I can merge this we need to understand what happens if this is installed on a system without snapd2.62 (ie one without `lxd-support-with-unconfined-mode` support)?

The LXD snap needs operate on older Ubuntu releases that don't necessarily have this snapd version.

Please can you advise about the backwards compatibility of this change?

alexmurray:
The `assumes: snapd2.62` means it can not be installed on a system without snapd 2.62 - so there is no risk here.

tomponline:
> The `assumes: snapd2.62` means it can not be installed on a system without snapd 2.62 - so there is no risk here.

Hrm, that sounds like it could be quite risky though - we would need to understand what happens on systems that do not yet have snapd 2.62 or where its pinned to a previous version.

tomponline:
There are multiple series of LXD:

5.0 - Previous LTS that needs to be able to run on >= Jammy (and ideally before that to allow upgrades).
5.21 - Current LTS, at minimal Noble, but ideally before that to allow upgrades.
latest - Currently the same as 5.21 but we can have higher release restrictions here - but we need to understand them well so we don't accidentally backport something to the LTS series that we shouldn't do.

So starting off we should consider the 5.0 series and what needs to be included in it to allow it to run with these restrictions and whether that will work on Jammy onwards. Also considering that LXD is preseeded into the Jammy images and so the snapd refresh may not have run on first LXD use.

tomponline:
@alexmurray shall we have a meeting about this to discuss?

alexmurray:
snapd 2.62 just migrated into the updates pocket for all the stable releases of Ubuntu and has been in the snapd snap stable channel for a while now too. So it should be quite safe to merge this PR now.

tomponline:
Thanks! We can try this now on various systems and check it doesn't cause a regression.

tomponline:
@mihalicyn this pr can help with your roadmap item to support restricted user namespaces.

mihalicyn:
I think we can merge this today, so we can test if there are any problems with it.

For me it looks perfectly safe and valid.

We have merged two other PRs related to this:
https://github.com/canonical/lxd/pull/12713
https://github.com/canonical/lxd-pkg-snap/pull/189

which are preparation to having this change if I understand correctly.

cc @tomponline

tomponline (approved):

tomponline:
Lets give it a go - this will need to be tested on bionic onwards please.

tomponline:
@alexmurray @mihalicyn this change is causing lxd snaps to be rejected by the snap store, so they are building ok in launchpad, but not being accepted for upload. Effectively this is now blocking latest/edge and will need to be reverted.

https://dashboard.snapcraft.io/snaps/lxd/revisions/29132/

> unknown attribute 'enable-unconfined-mode' for interface 'lxd-support' (plugs) lint-snap-v2_plugs_attributes (lxd-support-with-unconfined-mode, enable-unconfined-mode) 

@alexmurray please can you advise what the next steps are here?

mihalicyn:
Theoretically, this should allow using this option in snap store https://git.launchpad.net/review-tools/commit/reviewtools/sr_common.py?id=08e83bf8d0b36eb8ff82ae74e774f00d56493d5f

alexmurray:
@tomponline we are waiting on the store team to deploy the latest review-tools revision to the store dashboard - in the meantime we can manually approve revisions which I will do now - and I will poke the store team to get the latest review-tools deployed.

tomponline:
> in the meantime we can manually approve revisions which I will do now

How do we do that?

alexmurray:
Unfortunately it's a manual process to approve each revision, apologies for the delay. They are now published.

tomponline:
Do you know how long it will take to get the snapstore updated? We may need to revert this change if its going to take a while.

roosterfish:
Hi, I am seeing some permission errors in the snap. Maybe related to those changes here?

When trying to force stop an instance in error state using `lxc stop -f v1` I get:
`Error: Failed to stop VM process 227032: permission denied`

Kernel log shows this apparmor denied message:
`[33170.368181] audit: type=1400 audit(1719850282.401:24245): apparmor="DENIED" operation="signal" class="signal" profile="lxd-v1_</var/snap/lxd/common/lxd>" pid=227614 comm="lxd" requested_mask="receive" denied_mask="receive" signal=kill peer="snap.lxd.daemon"`

It's reproducible by repeating this https://github.com/canonical/lxd-ci/blob/main/tests/storage-vm#L226

mihalicyn:
@roosterfish thanks for reporting this, Julian!

Related to:
https://bugs.launchpad.net/apparmor/+bug/2067900

simondeziel:
> snapd 2.62 just migrated into the updates pocket for all the stable releases of Ubuntu and has been in the snapd snap stable channel for a while now too. So it should be quite safe to merge this PR now.

Just a quick note (at least for future me). This doesn't work as `snapd` doesn't self-refresh when meeting a snap with a higher required version:

```
+ snap refresh lxd --channel=latest/edge --cohort=+
error: cannot refresh "lxd": snap "lxd" assumes unsupported features: snapd2.64
       (try to refresh snapd)
```

This ^ was on machine that was still running with `snapd` 2.63, either because it had not applied updates or was configured with only `-security` enabled. At the time of this failure:

```
$ rmadison snapd
 snapd | 2.0.2               | xenial           | source, amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 snapd | 2.32.5+18.04        | bionic           | source, amd64, arm64, armhf, i386, ppc64el, s390x
 snapd | 2.37.4~14.04.1      | trusty-security  | source, amd64, armhf, i386
 snapd | 2.38~14.04          | trusty-updates   | source, amd64, armhf, i386
 snapd | 2.44.3+20.04        | focal            | source, amd64, arm64, armhf, ppc64el, s390x
 snapd | 2.48.3              | xenial-security  | source, amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 snapd | 2.48.3              | xenial-updates   | source, amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 snapd | 2.55.3+22.04        | jammy            | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.58+18.04.1        | bionic-security  | source, amd64, arm64, armhf, i386, ppc64el, s390x
 snapd | 2.58+18.04.1        | bionic-updates   | source, amd64, arm64, armhf, i386, ppc64el, s390x
 snapd | 2.62+24.04build1    | noble            | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.63+20.04ubuntu0.1 | focal-security   | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.63+22.04ubuntu0.1 | jammy-security   | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.63+24.04ubuntu0.1 | noble-security   | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.65.3+24.10        | oracular         | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.66.1+20.04        | focal-updates    | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.66.1+22.04        | jammy-updates    | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.66.1+24.04        | noble-updates    | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.66.1+24.10        | oracular-updates | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 snapd | 2.66.1+25.04        | plucky           | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
```

```
$ snap info snapd
name:      snapd
summary:   Daemon and tooling that enable snap packages
publisher: Canonical✓
store-url: https://snapcraft.io/snapd
contact:   https://github.com/snapcore/snapd/issues
license:   GPL-3.0
description: |
  Install, configure, refresh and remove snap packages. Snaps are
  'universal' packages that work across many different Linux systems,
  enabling secure distribution of the latest apps and utilities for
  cloud, servers, desktops and the internet of things.
  
  Start with 'snap list' to see installed snaps.
type:         snapd
snap-id:      PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4
tracking:     latest/stable
refresh-date: 19 days ago, at 17:18 EST
channels:
  latest/stable:    2.67                2025-01-11 (23545) 46MB -
  latest/candidate: 2.67                2024-12-17 (23545) 46MB -
  latest/beta:      2.67.1              2025-01-17 (23771) 46MB -
  latest/edge:      2.67.1+g190.9d4a716 2025-01-28 (23853) 50MB -
installed:          2.67                           (23545) 46MB snapd
```

PR #278, opened 2024-01-24
snapcraft/wrappers/run-host: use unshare --root instead of calling chroot

simondeziel:
Both are equivalent and we already use unshare --root elsewhere so one less moving part and more consistency.


simondeziel:
This was tested with `EDITOR=vim lxc config edit` and also by using `lvm.external=true` while having disabled the `nsenter` path from `run-host` so that is uses the `unshare` path.

tomponline (approved):

PR #285, opened 2024-01-25
snapcraft: strip criu binary (5.9M -> 1.6M)

simondeziel:
`criu` wasn't being stripped because it is not under `bin/` due to being optionally added to $PATH when `snap set lxd criu.enable=true`.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1591,6 +1591,10 @@ parts:
         -not -path "${CRAFT_PRIME}/bin/xfs_admin" \
         -exec strip -s {} +
 
+      # Strip binaries not under bin/ due to being dynamically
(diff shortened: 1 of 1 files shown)


tomponline (approved):

simondeziel:
Merging as locally built snap confirmed that it's now properly stripped:

```
# file /snap/lxd/current/criu/criu 
/snap/lxd/current/criu/criu: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=8666cb872be18d8fdbc8be758f275559a0b4d1c3, for GNU/Linux 3.2.0, stripped
```

PR #286, opened 2024-01-25
Faster snap builds

simondeziel:
This was tested on amd64 only where the speedup is minimal as the effective parts list is the same.

For arches for which we don't build QEMU (like the slow riscv64), it should be more visible as the `nasm`, `spice-protocol` and `spice-server` parts are now skipped.

Note: while making the "architecture lists" for skipping parts, I noticed there are 2 lists the QEMU-enabled list and the CRIU-enabled one. The CRIU one is the same as the QEMU except for armhf. That's OK, just something that caught my eye.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -766,6 +766,13 @@ parts:
     stage-packages:
       - libjpeg-turbo8
       - libpixman-1-0
+    override-pull: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      craftctl default
+
+    override-build: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      craftctl default
     organize:
       sbin/: bin/
       usr/lib: lib/
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -407,6 +407,7 @@ parts:
 
   libmnl:
     source: https://git.netfilter.org/libmnl
+    source-depth: 1
     source-tag: libmnl-1.0.5
     source-type: git
     plugin: autotools
@@ -421,6 +422,7 @@ parts:
     after:
       - libmnl
     source: https://git.netfilter.org/libnftnl
+    source-depth: 1
     source-tag: libnftnl-1.2.6
     source-type: git
     plugin: autotools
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -256,7 +256,7 @@ parts:
         [ "$(uname -m)" != "ppc64le" ] && exit 0
       set -ex
 
-      make USERCFLAGS=-Wno-error=format-truncation
+      make USERCFLAGS=-Wno-error=format-truncation criu
```
tomponline, line 251: what else was being built here?

snapcraft.yaml
```diff
@@ -296,12 +296,8 @@ parts:
     plugin: nil
     build-packages:
       - g++
-      - on amd64:
-        - acpica-tools
-        - uuid-dev
-      - on arm64:
-        - acpica-tools
-        - uuid-dev
+      - acpica-tools
```
tomponline, line 291: wont this cause these packages to be unnecessarily installed on some archs?

tomponline:
@simondeziel can you also add a fix for https://github.com/canonical/lxd-pkg-snap/pull/285 in this PR as its failing on riscv

```
:: strip: '/build/lxd/prime/criu/criu': No such file
```

https://launchpadlibrarian.net/711148992/buildlog_snap_ubuntu_jammy_riscv64_lxd-latest-edge_BUILDING.txt.gz

snapcraft.yaml
simondeziel: https://github.com/checkpoint-restore/criu/blob/criu-dev/Makefile#L167-L168:

```
# Default target
all: flog criu lib crit
```

At first, the `lib` one might seem interesting but looking at criu's prime section, the only 2 libs it primes are coming from the stage-packages `libnet1` and `libprotobuf-c1`.

snapcraft.yaml
simondeziel: Those are in the build-packages section so presumably they will be installed on all arches (same as the `g++` one) but after looking at https://packages.ubuntu.com/focal/acpica-tools, it seems it was not available on `riscv64` so probably why this was made arch-dependent.

Now this is no longer applicable on core22 as all 3 packages are available on all arches. So they will be installed but the part as a whole will no do anything for some arches. Those packages' bits won't end up in the final snap as they are in the build-packages section.

So I think this is OK as it is, LP builders might prove me wrong though.

tomponline: mmm ok cool.

tomponline (approved):

snapcraft.yaml
simondeziel: With the above said, some time could be saved by (re-)introducing the `- on amd64,arm64` bit in the build-packages so we skip the installation of those 3 packages for say riscv64.

Will do that in a follow-up PR, thanks for bringing that up!

PR #287, opened 2024-01-26
snapcraft: stop priming nasm

simondeziel:
For reasons unclear to me, the nasm binary was still in the end snap despite not being primed anymore. I can only guess that since edk2 used it, snapcraft assumed it was needed somehow? Anyhow, it's not being removed during the `strip` part.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -582,8 +582,6 @@ parts:
       - --prefix=
     organize:
       usr/bin/: bin/
-    prime:
-      - bin/nasm
     override-pull: |-
       [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
       craftctl default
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -582,8 +582,6 @@ parts:
       - --prefix=
     organize:
       usr/bin/: bin/
-    prime:
-      - bin/nasm
```
tomponline, line 586: im guessing if there's no prime section it copies everything from staging

tomponline (approved):

snapcraft.yaml
simondeziel: Oh that could be, let me check if it doesn't bring uneeded bits then.

simondeziel: Right on the money as always. I'm working on a cleanup PR

PR #299, opened 2024-02-01
snapcraft: Change description of ui.enable (latest-candidate)

tomponline:
UI is no longer preview/experimental.

Signed-off-by: Thomas Parrott <thomas.parrott@canonical.com>
(cherry picked from commit 038295972cd10805e3142993b597397e2f00bb27)

And pin go version.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -48,7 +48,7 @@ description: |-
    - openvswitch.builtin: Run a snap-specific OVS daemon [default=false]
    - openvswitch.external: Use the system's OVS tools (ignores openvswitch.builtin) [default=false]
    - ovn.builtin: Use snap-specific OVN configuration [default=false]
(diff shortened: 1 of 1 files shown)


simondeziel (dismissed):

simondeziel (approved):
Not sure why go needed pining though.

tomponline:
> Not sure why the go needed pining though.

Imagine a scenario where we come to do a cherry-pick later on and latest/stable channel of the go snap is using go 1.22 (for example) then LXD will be rebuilt with a different go version than when we originally published the last version.

This may result in unexpected behavioral changes (for example Go 1.22 has some subtle but important changes to for loop var behaviors).

For "*/candidate" builds I want them to be as predictable and as reproducible as possible so we dont end up accidentally introducing unexpected changes in later rebuilds.

PR #301, opened 2024-02-04
snapcraft: make it clear that lz4 lib isn't primed

simondeziel:
snapcraft deduplication is not applied to files from packages listed in
stage-packages.  To avoid duplicating lz4's .so, make it explicit that we don't
want to prime it and instead relying on the one provided in the base snap.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -978,6 +978,7 @@ parts:
     prime:
       - lib/libraft*so*
       - lib/*/libuv.so*
+      #- lib/*/liblz4.so*  # use liblz4.so from the base snap
 
   sqlite:
     source: https://github.com/sqlite/sqlite


snapcraft.yaml
```diff
@@ -978,6 +978,7 @@ parts:
     prime:
       - lib/libraft*so*
       - lib/*/libuv.so*
+      #- lib/*/liblz4.so*  # use liblz4.so from the base snap
```
tomponline, line 981: @simondeziel out of interest, how do we know what is provided in the base snap (assume you mean core22)?

tomponline (approved):

snapcraft.yaml
simondeziel: @tomponline the core snaps all contain `/snap/<base>/current/usr/share/snappy/dpkg.list` as mentioned in https://snapcraft.io/docs/build-and-staging-dependencies#heading--filtering

Now for lz4 specifically, I checked `core18`,  `core20`, `core22` and `core24` and they all bundle it. I don't know if that's somehow a dependency due to snaps being potentially lz4-compressed?

PR #302, opened 2024-02-04
Minimize priming

simondeziel:


snapcraft.yaml
```diff
@@ -1359,11 +1359,9 @@ parts:
       - gdisk
       - iw
       - lshw
-      - netbase
       - pciutils
       - pigz
       - rsync
-      - squashfs-tools
```
tomponline, line 1365: is the tooling still included in the snap though?

tomponline (approved):

snapcraft.yaml
simondeziel: It never made it in our snap as nothing primed it. The tools either come from `squashfs-tools` that is in the base snap or our vendored `squashfs-tools-ng`.

tomponline (approved):

PR #305, opened 2024-02-05
snapcraft: use libseccomp from the base snap

simondeziel:
In 2019, when commit cddfdfc76f4d46a0 was added, LXD used core (16) as base which probably came with a rather old libseccomp not adequate for modern kernels at that time (4.15+).  Now that we are on core22, we get 2.5.3 instead of pulling 2.5.5 from upstream. With core24, we will get 2.5.4 or maybe newer.

For some reason, virtiofsd used the libseccomp-dev package despite having a newer lib shipped in the snap.

With the intention of using QEMU from deb sources, using the same libseccomp version makes even more sense.


simondeziel:
FYI, this was tested with a locally built snap and then running (nested) VMs and seeing QEMU/virtiofsd work.

tomponline (approved):

PR #308, opened 2024-02-06
Remove doc build artifacts

simondeziel:
The files/dirs being removed:

```
$ du -sc /snap/lxd/current/share/lxd-documentation/{.buildinfo,objects.inv,_sphinx_design_static/,_sources/}
1	/snap/lxd/current/share/lxd-documentation/.buildinfo
18	/snap/lxd/current/share/lxd-documentation/objects.inv
49	/snap/lxd/current/share/lxd-documentation/_sphinx_design_static/
1162	/snap/lxd/current/share/lxd-documentation/_sources/
1229	total
```

Without the above, the UI and the builtin doc still worked fine.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1413,7 +1413,7 @@ parts:
         rm -rf doc/html/_sphinx_design_static/
         rm -rf doc/html/_sources/
 
-        # Copy the static website
+        # Stage the static website
         mkdir -p "${CRAFT_STAGE}/share/lxd-documentation"
         cp -a doc/html/. "${CRAFT_STAGE}/share/lxd-documentation/"
       fi
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1406,6 +1406,13 @@ parts:
         # Build the static website
         make doc
 
+        # Remove unneeded bits
+        rm doc/html/objects.inv    # only objects.inv.txt is used
+        # not needed once built
+        rm doc/html/.buildinfo
+        rm -rf doc/html/_sphinx_design_static/
+        rm -rf doc/html/_sources/
+
         # Copy the static website
         mkdir -p "${CRAFT_STAGE}/share/lxd-documentation"
         cp -a doc/html/. "${CRAFT_STAGE}/share/lxd-documentation/"


snapcraft.yaml
```diff
@@ -1406,7 +1406,14 @@ parts:
         # Build the static website
         make doc
 
-        # Copy the static website
+        # Remove unneeded bits
+        rm doc/html/objects.inv    # only objects.inv.txt is used
```
tomponline, line 1410: @ru-fu @gabrielmougard is this right? Do we  only need the .txt file for exposing to the UI?

ru-fu: I guess that's a question for @edlerd , but as far as I know the UI is using only the text file.

The `objects.inv` file would be needed when linking between different doc sets - so if we wanted to have interlinked local documentation for, for example, MicroCloud and LXD in the future, we would still need it. But we don't have that at the moment.

tomponline (approved):

snapcraft.yaml
```diff
@@ -1406,7 +1406,14 @@ parts:
...
+        # not needed once built
+        rm doc/html/.buildinfo
+        rm -rf doc/html/_sphinx_design_static/
+        rm -rf doc/html/_sources/
```
ru-fu, line 1414: The sources are linked at the bottom of each page, so this will cause broken links in the footer.

tomponline: @simondeziel ah dam, please can you revert.

ru-fu: Alternatively, we could probably find some way to use a different footer for snap builds. Not sure if the sources are needed for everyone (I would like to keep them for the online docs and normal local build though, since they are very useful when working with the docs).

tomponline: Can you link to the online ones always?

ru-fu: This would be error-prone since the doc version shipped with the snap might differ from what is online - so the sources might not correspond to the documentation shipped with the snap.

tomponline: So you're thinking just removing the link for offline version?

ru-fu: Yes - or more precisely, for the docs that are shipped with the snap. We can control this by setting some environment variable for the snap build.

tomponline: got it, makes sense, thanks

snapcraft.yaml
simondeziel: Only the `objects.inv.txt` file is being used it seems https://github.com/search?q=repo%3Acanonical%2Flxd-ui%20objects.inv&type=code

snapcraft.yaml
simondeziel: Thanks for catching this @ru-fu. #313 stops removing those sources.

PR #311, opened 2024-02-08
Remove minio

masnax:
Removes `minio` so that it is no longer included with the LXD snap.

Introduces a `minio.path` snap config key which can be user specified to point to the locally installed `minio`. This will get included in `PATH` by the snap. One thing to note is that the snap won't work with just any path, for example the snap won't allow executing a binary from a path that includes a user's home directory.

---

~~Currently, LXD calls `minio` directly from `PATH`. I wanted to preserve this so that we don't need to have special cases depending on the method of installation of LXD (snap or otherwise).~~

~~I first tried to just symlink `/var/lib/snapd/hostfs/usr/bin/minio` to `/usr/bin/minio` in the snap, but this doesn't work because `/usr` is read-only. It's also limiting in case the user has `minio` in their `PATH` but not in `/usr/bin`.~~

~~So instead, I opted to iterate over `PATH` in the snap confinement, ignoring any snap-specific directories, and checking if an executable `minio` exists in the equivalent directory on the hostfs. If so, then it's symlinked to `/var/snap/lxd/common/bin` which is included in LXD's `PATH` now.~~

~~This means that if LXD is installed via the snap, `minio` will be accessible from any of the following paths, and symlinked to `/var/snap/lxd/common/bin`:~~
```
  /usr/local/sbin
  /usr/local/bin
  /usr/sbin
  /usr/bin
  /sbin
  /bin
  /usr/games
  /usr/local/games
```

~~One note is that I see `/run/bin` is already a custom directory included in `PATH` in the snap. I wasn't sure if I should re-use this directory or not, and instead opted to add a new directory at `/var/snap/lxd/common/bin` which is a bit more visible.~~

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -606,6 +606,13 @@ if [ -x "${SNAP_COMMON}/lxd.debug" ]; then
     echo "==> WARNING: Using a custom debug LXD binary!"
 fi
 
+# Note: Snaps disallow running binaries from certain paths (for example paths under home directories).
+# These will show `permission denied` when trying to run the executable.
+if [ -n "${minio_path}" ] ; then
+  minio_dir="$(dirname "/var/lib/snapd/hostfs${minio_path}")"
+  export PATH="${PATH}:${minio_dir}"
+fi
+
 CMD="${LXD} --logfile ${SNAP_COMMON}/lxd/logs/lxd.log"
 
 if getent group "${daemon_group}" >/dev/null 2>&1; then
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -525,36 +525,6 @@ parts:
       - lib/*/liblvm*
       - lib/*/libreadline.so*
 
-  minio:
-    source: https://github.com/minio/minio
-    source-depth: 1
-    source-tag: RELEASE.2023-11-20T22-40-07Z
-    source-type: git
-    plugin: nil
-    build-snaps:
-      - go
-    override-prime: |-
-      [ "$(uname -m)" = "riscv64" ] && exit 0
-      craftctl default
-    override-pull: |-
-      [ "$(uname -m)" = "riscv64" ] && exit 0
-      craftctl default
-    override-build: |-
-      [ "$(uname -m)" = "riscv64" ] && exit 0
-      set -ex
-
-      # Setup build environment
-      export GOPATH="$(realpath ./.go)"
-
-      # Build the binaries
-      make build
-
-      # Install minio command
-      mkdir -p "${CRAFT_PART_INSTALL}/bin/"
-      cp minio "${CRAFT_PART_INSTALL}/bin/minio"
-    prime:
-      - bin/minio*
-
   nano:
     plugin: nil
     stage-packages:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -49,6 +49,7 @@ description: |-
    - openvswitch.external: Use the system's OVS tools (ignores openvswitch.builtin) [default=false]
    - ovn.builtin: Use snap-specific OVN configuration [default=false]
    - ui.enable: Enable the web interface [default=true]
+   - minio.path: Path to the minio binary to use with LXD [default=""]
 
  For system-wide configuration of the CLI, place your configuration in
  /var/snap/lxd/common/global-conf/ (config.yml and servercerts)
--- a/snapcraft/hooks/configure
+++ b/snapcraft/hooks/configure
@@ -56,6 +56,7 @@ openvswitch_builtin=$(get_bool "$(snapctl get openvswitch.builtin)")
 openvswitch_external=$(get_bool "$(snapctl get openvswitch.external)")
 ovn_builtin=$(get_bool "$(snapctl get ovn.builtin)")
 ui_enable=$(get_bool "$(snapctl get ui.enable)")
+minio_path="$(snapctl get minio.path)"
 
 # Special-handling of daemon.preseed
 daemon_preseed=$(snapctl get daemon.preseed)
@@ -85,6 +86,7 @@ config="${SNAP_COMMON}/config"
     echo "openvswitch_external=${openvswitch_external:-"false"}"
     echo "ovn_builtin=${ovn_builtin:-"false"}"
     echo "ui_enable=${ui_enable:-"true"}"
+    echo "minio_path=${minio_path:-}"
 } > "${config}"
 
 # Set socket ownership in case it changed


tomponline (commented):
I was thinking about this the other night and I'm thinking we should introduce  snap package config called `minio.path` that in turn runs LXD with a `LXD_MINIO_PATH` env var which we will then use in LXD to run minio rather than searching for it in the PATH env.

This would make it explicit that you have to enable an external minio binary, and would allow the minio binary to be anywhere the admin chooses.

What do you think?

masnax:
> I was thinking about this the other night and I'm thinking we should introduce snap package config called `minio.path` that in turn runs LXD with a `LXD_MINIO_PATH` env var which we will then use in LXD to run minio rather than searching for it in the PATH env.
> 
> This would make it explicit that you have to enable an external minio binary, and would allow the minio binary to be anywhere the admin chooses.
> 
> What do you think?

This will mean users would have to set a snap config to the path of their minio, which gets set to the environment variable by the snap itself, right? Out of curiosity, why do you prefer to set this at the snap level and set an environment var rather than just directly setting the path via a LXD config key? In the latter there's no need to document the 2 ways to use minio. (Environment variable without snap, and snap config with snap)

tomponline:
> This will mean users would have to set a snap config to the path of their minio, which gets set to the environment variable by the snap itself, right? Out of curiosity, why do you prefer to set this at the snap level and set an environment var rather than just directly setting the path via a LXD config key?

Yes I would prefer a snap config variable, which could then actually be appended to the PATH that LXD runs with (if thats possible) so we dont need snap or minio specific logic in LXD.

After all this change is only to do with the problems of distributing minio in the snap and shouldnt affect the LXD source code itself.

masnax:
> > This will mean users would have to set a snap config to the path of their minio, which gets set to the environment variable by the snap itself, right? Out of curiosity, why do you prefer to set this at the snap level and set an environment var rather than just directly setting the path via a LXD config key?
> 
> Yes I would prefer a snap config variable, which could then actually be appended to the PATH that LXD runs with (if thats possible) so we dont need snap or minio specific logic in LXD.
> 
> After all this change is only to do with the problems of distributing minio in the snap and shouldnt affect the LXD source code itself.

Well LXD would still need to get updated to read LXD_MINIO_PATH which is a change to the LXD source code.

Additionally, for the case of LXD being installed without a snap, we would have to document that you have to set LXD_MINIO_PATH, but if you use the snap this won't work and you'll have to set a snap config key instead.

tomponline:
> > > This will mean users would have to set a snap config to the path of their minio, which gets set to the environment variable by the snap itself, right? Out of curiosity, why do you prefer to set this at the snap level and set an environment var rather than just directly setting the path via a LXD config key?
> > 
> > 
> > Yes I would prefer a snap config variable, which could then actually be appended to the PATH that LXD runs with (if thats possible) so we dont need snap or minio specific logic in LXD.
> > After all this change is only to do with the problems of distributing minio in the snap and shouldnt affect the LXD source code itself.
> 
> Well LXD would still need to get updated to read LXD_MINIO_PATH which is a change to the LXD source code.
> 
> Additionally, for the case of LXD being installed without a snap, we would have to document that you have to set LXD_MINIO_PATH, but if you use the snap this won't work and you'll have to set a snap config key instead.

Hence why i asked:

> Yes I would prefer a snap config variable, which could then actually be appended to the PATH that LXD runs with (if thats possible) so we dont need snap or minio specific logic in LXD.

masnax:
Let me rephrase my question:

A: With a LXD config key:
 * If installed without a snap LXD would need to try the path at `LXD_MINIO_PATH`.
 * If installed with a snap, it will need to read `/var/lib/snapd/hostfs/${LXD_MINIO_PATH}` 
This means introducing special cases to the LXD source code which is bad.

B: With a pure environment variable:
 * If installed without a snap, we need to document LXD_MINIO_PATH to be set by the user.
 * If installed with a snap, we need to document `minio.path` to be set by the user instead.
This means documenting a special case rather than keeping it in the source code.

What makes the outcome of option B preferable to option A? (documenting a special case vs keeping it in the code).

masnax:
> Yes I would prefer a snap config variable, which could then actually be appended to the PATH that LXD runs with (if thats possible) so we dont need snap or minio specific logic in LXD.


Perhaps I was confused by the wording, are you trying to say we shouldn't use `LXD_MINIO_PATH` from your first commentand instead just use a snap config with `PATH` itself? I'm happy with that.

tomponline:
Yes if we can add it to normal PATH variable that LXD is invoked with we wont need to make a change to LXD's code.

masnax:
@tomponline Just as an FYI, when testing the `minio.path` approach, it appears that the snap won't necessarily accept just any path that the user specifies. For example it can't be under a user's home directory or running minio will return `permission denied`.

simondeziel (dismissed):

snapcraft/commands/daemon.start
```diff
@@ -606,6 +606,15 @@ if [ -x "${SNAP_COMMON}/lxd.debug" ]; then
     echo "==> WARNING: Using a custom debug LXD binary!"
 fi
 
+full_minio_path="/var/lib/snapd/hostfs${minio_path}"
+# Ensure minio_path exists and is an executable file on the hostfs.
+# Note: Snaps disallow running binaries from certain paths (for example paths under home directories).
+# These will show `permission denied` when trying to run the executable.
+if [ -n "${minio_path}" ] && [ -f "${full_minio_path}" ] && [ -x "${full_minio_path}" ]; then
+  minio_dir="$(dirname "${full_minio_path}")"
+  export PATH="${PATH}:${minio_dir}"
```
tomponline, line 613: Could we just add it to the path if non empty and let lxd and go resolve it and check for being executable like it normally does now?

masnax: Done.

simondeziel (commented):
Is minio executed as root? If so, are there any concerns with allowing a user to provide the path to it?

tomponline (approved):

tomponline:
@masnax please can you update the tests over at https://github.com/canonical/lxd-ci/blob/main/tests/storage-buckets to download minio on the system (and remove it at the end) and set the necessary snap config settings. Thanks

tomponline:
Otherwise that test will be failing now.

simondeziel:
I'm fine with losing minio from the snap but how are we going to warn the users about that change? I'm assuming an API extension (`bundled_minio` or something) is out of question, right?

So just the announcement with the next LXD release?

tomponline:
> So just the announcement with the next LXD release?

Yep ill make sure its predominantly displayed, and as its a new LTS series its a good (only) time to do it really.

Its only affecting snap packaging so not suitable for an API extension.

simondeziel:
Thanks Tom.

So since it's not discoverable through the API, the lxd-ci change should probably peak at `/snap/lxd/current/bin/minio` to decide if minio needs to be downloaded/cleaned.

tomponline:
sounds good

PR #323, opened 2024-02-09
Provide default value for minio.path

simondeziel:


masnax (approved):
Yikes, that you need quotation marks like that to get it to work is really upsetting to me, but it is what it is :)

simondeziel:
I don't know if the quotation marks are strictly required to be honest.

tomponline (approved):

PR #324, opened 2024-02-12
snapcraft: add python-uefivars

mihalicyn:
We need it for "lxc config uefi <...>" extension

There is some trickiness around this thing, because for some reason snapcraft is going crazy and everything that builds after "uefivars" starts to fail with some Python errors (failed to find package XYZ).

I have workarounded it by adding "after" specifier for uefivars part.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -602,6 +602,7 @@ parts:
     autotools-configure-parameters:
       - --prefix=
       - --with-json
+      - --disable-man-doc
     build-packages:
       - libedit-dev
       - libjansson-dev
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1378,6 +1378,17 @@ parts:
 
       patch -p1 "${CRAFT_PART_INSTALL}/snap/lxd/current/lxcfs/lxc.mount.hook" < "${CRAFT_PROJECT_DIR}/patches/lxcfs-0001-hook.patch"
 
+  uefivars:
+    after:
+      - spice-server
+      - qemu-ovmf-secureboot
+      - nftables
+    source: https://github.com/awslabs/python-uefivars
+    source-depth: 1
+    source-tag: v1.0.0
+    source-type: git
+    plugin: python
+
   lxd:
     source: https://github.com/canonical/lxd
     source-depth: 1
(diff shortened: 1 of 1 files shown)


mihalicyn:
I have no posted a PR for LXD yet, but I'll do that in next few days. It's safe to merge this before that as here we are just adding a new component.

snapcraft.yaml
```diff
@@ -1375,6 +1376,16 @@ parts:
 
       patch -p1 "${CRAFT_PART_INSTALL}/snap/lxd/current/lxcfs/lxc.mount.hook" < "${CRAFT_PROJECT_DIR}/patches/lxcfs-0001-hook.patch"
 
+  uefivars:
```
MusicDin, line 1381: I see there is only one release of python-uefivars - but should we bind it to the specific tag (`1.0.0`), or is latest ok?

tomponline: we should indeed or a git sha

simondeziel (commented):
Agreed on the version pinning suggested by Din.

As for this new part, it should be noted that once on core24, we will be able to use the [`python3-uefivars`](https://packages.ubuntu.com/noble/python3-pyuefivars) package instead :)

snapcraft.yaml
mihalicyn: Sure, I'll do version pinning. My bad. Thanks!

snapcraft.yaml
```diff
@@ -600,6 +600,7 @@ parts:
     autotools-configure-parameters:
       - --prefix=
       - --with-json
+      - --disable-man-doc
```
tomponline, line 605: This should be a separate commit to make back porting easier please

snapcraft.yaml
tomponline: @mihalicyn please can you do the version pinning and i can merge this, ta

snapcraft.yaml
mihalicyn: Have done!

snapcraft.yaml
mihalicyn: Have done!

tomponline (approved):
Thanks!

PR #325, opened 2024-02-12
snapcraft: edk2: update logo

mihalicyn:
```
Converted from "LXD tags@2x.png" using:
$ convert -scale 60% -colors 256 -depth 8 -type palette -compress none LXD\ tags@2x.png BMP3:edk2-0002-logo.bmp
```


tomponline:
@mihalicyn does it work with a rectangular logo or do we need to crop it to be square?

mihalicyn:
> @mihalicyn does it work with a rectangular logo or do we need to crop it to be square?

The previous image also wasn't a perfect square. But right now I'm testing this change together with the edk2 version update ;-)

tomponline:
> > @mihalicyn does it work with a rectangular logo or do we need to crop it to be square?
> 
> The previous image also wasn't a perfect square. But right now I'm testing this change together with the edk2 version update ;-)

If it looks OK im happy with using he image as it is.

mihalicyn:
It works perfectly fine now. One extra fix was needed for the image.
I have added a `-compress none` as EDKII does not support compressed bmp files.

mihalicyn:
Safe to merge.

tomponline (approved):

PR #326, opened 2024-02-12
snapcraft: edk2: update to stable202311

mihalicyn:
Have tested it with Ubuntu Bionic and it boots just fine. We still need to have all the custom workarounds for older kernels.


mihalicyn:
Tested working on amd64 machine. I will do some extensive testing on arm64 after. I think we can merge it to `latest-edge` to simplify testing on arm.

tomponline (approved):

PR #175, opened 2023-10-03
Disable Apparmor unprivileged userns mediation

simondeziel:
Implement the short-term solution for [#11920](https://github.com/canonical/lxd/issues/11920)

Tests done:

The snap built from this PR effectively disables the AA mediation if need be:
```
root@ma:~# echo 1 > /proc/sys/kernel/apparmor_restrict_unprivileged_userns
root@ma:~# snap start lxd
Started.
root@ma:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
0
root@ma:~# journalctl -b0 -u snap.lxd.daemon.service --grep Apparmor
Oct 03 16:13:27 ma lxd.daemon[7351]: ==> Disabling Apparmor unprivileged userns mediation
```

And this allows using the EDITOR:

```
root@ma:~# export EDITOR=vim
root@ma:~# lxc config edit c1
```

```
root@ma:~# ps faux
...
root         336  0.1  1.8 1240108 17668 ?       Ssl  16:06   0:01 /run/lxd_agent/lxd-agent
root        1222  0.0  0.3   9696  3072 pts/0    Ss   16:06   0:00  \_ su -l
root        1246  0.0  0.4   9152  4480 pts/0    S    16:06   0:00  |   \_ -bash
root        4460  0.0  0.6  34852  6272 pts/0    S    16:09   0:00  |       \_ journalctl -fu snap.lxd.daemon
root       11268  1.2  1.7 1538916 17152 pts/0   Sl+  16:19   0:00  |       \_ /snap/lxd/x2/bin/lxc config edit c1
root       11289  0.0  0.1   3208  1664 pts/0    S+   16:19   0:00  |           \_ unshare --kill-child -U -m -p -r -f -R /var/lib/snapd/hostfs/ /bin/sh -c mount -t proc proc /proc 2>/dev/null || true; exec "vim" "/proc/self/fd/9/lxd_editor_1580859729.yaml"
root       11295  2.1  1.3  24244 12928 pts/0    Sl+  16:19   0:00  |               \_ vim /proc/self/fd/9/lxd_editor_1580859729.yaml
```

If one re-enables the AA mediation after LXD started, it will use `vim.tiny` from the snap:

```
root@ma:~# echo 1 > /proc/sys/kernel/apparmor_restrict_unprivileged_userns
root@ma:~# export EDITOR=vim
root@ma:~# lxc config edit c1
```

```
root@ma:~# ps faux
...
[...]

PR #327, opened 2024-02-13
Apparmor unprivileged restrictions disable

simondeziel:
At the distro level, Ubuntu 24.04 LTS should ship with additional restrictions preventing unconfined processes from creating userns by default.

The long term goal is for LXD to play nice with those additional restrictions but since a lots of moving parts are involved (kernel, kernel config, apparmor parser, core24 snap, snapd and LXD itself) and not fully ready yet, we decided on disabling those restrictions globally by default in the short term. This decision also took into consideration that LXD needs to work across many distros and versions (Debian w/ Apparmor, old Ubuntu version, old Ubuntu version with HWE kernel, etc).

This PR thus introduce a new `snap config` key (`apparmor.unprivileged-restrictions-disable`) that in its default value (`true`) causes LXD to set `kernel.apparmor_restrict_unprivileged_userns=0` and `kernel.apparmor_restrict_unprivileged_unconfined=0` thus disabling the new restrictions.

For additional information on those new restrictions, see https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces and https://discourse.ubuntu.com/t/spec-unprivileged-user-namespace-restrictions-via-apparmor-in-ubuntu-23-10/37626.

This PR reintroduced bits of #175.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -45,11 +45,11 @@ description: |-
    - lxcfs.loadavg: Start tracking per-container load average [default=false]
    - lxcfs.cfs: Consider CPU shares for CPU usage [default=false]
    - lxcfs.debug: Increase logging to debug level [default=false]
+   - minio.path: Path to the minio binary to use with LXD [default=""]
    - openvswitch.builtin: Run a snap-specific OVS daemon [default=false]
    - openvswitch.external: Use the system's OVS tools (ignores openvswitch.builtin) [default=false]
    - ovn.builtin: Use snap-specific OVN configuration [default=false]
    - ui.enable: Enable the web interface [default=true]
-   - minio.path: Path to the minio binary to use with LXD [default=""]
 
  For system-wide configuration of the CLI, place your configuration in
  /var/snap/lxd/common/global-conf/ (config.yml and servercerts)
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -31,6 +31,7 @@ description: |-
 
  Supported configuration options for the snap (snap set lxd [<key>=<value>...]):
 
+   - apparmor.unprivileged-restrictions-disable: Whether to disable restrictions on unprivileged user namespaces [default=true]
    - ceph.builtin: Use snap-specific Ceph configuration [default=false]
    - ceph.external: Use the system's ceph tools (ignores ceph.builtin) [default=false]
    - criu.enable: Enable experimental live-migration support [default=false]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,39 @@
+name: Tests
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    # Make sure bash is always invoked with `-eo pipefail`
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
+    shell: bash
+
+jobs:
+  code-tests:
+    name: Code
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - id: ShellCheck
+        name: Differential ShellCheck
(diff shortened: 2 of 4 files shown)


simondeziel:
This was tested in a local 24.04 VM like this:

```
$ lxc launch ubuntu-daily:24.04 nv1 --vm
$ lxc shell nv1
root@nv1:~# snap install lxd   # getting the assertion in place
lxd 5.20-f3dd836 from Canonical✓ installed
root@nv1:~# snap install --dangerous /dev/shm/lxd_0+git.170c8352_amd64.snap 
lxd 0+git.170c8352 installed
root@nv1:~# snap alias lxd.lxc lxc
Added:
  - lxd.lxc as lxc
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
1
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_unconfined
0
```
So the keys have their default values before starting LXD's daemon.

```
root@nv1:~# systemctl start snap.lxd.daemon.service
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
0
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_unconfined
0
```
So LXD noticed that `apparmor_restrict_unprivileged_userns` needed to be disabled which it did as we can also confirm in the logs:

```
root@nv1:~# journalctl -u snap.lxd.daemon.service --grep apparmor
Feb 13 21:47:24 nv1 lxd.daemon[2325]: ==> Disabling Apparmor unprivileged userns mediation
```

And LXD behaves fine:

```
root@nv1:~# lxd init --auto
root@nv1:~# lxc launch ubuntu-minimal-daily:22.04 c1 -c security.nesting=true
root@nv1:~# lxc exec c1 -- hostname
c1
root@nv1:~# lxc exec c1 -- systemctl --failed
  UNIT LOAD ACTIVE SUB DESCRIPTION
0 loaded units listed.
root@nv1:~# 
```

And now to prevent LXD from disabling the Apparmor restriction:

```
root@nv1:~# snap set lxd apparmor.unprivileged-restrictions-disable=false
root@nv1:~# snap stop lxd
2024-02-13T21:57:45Z INFO Waiting for "snap.lxd.daemon.unix.socket", "snap.lxd.daemon.service" to stop.
Stopped.
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
0
root@nv1:~# echo 1 > /proc/sys/kernel/apparmor_restrict_unprivileged_userns
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
1
root@nv1:~# snap start lxd
Started.
root@nv1:~# cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns
1
```

simondeziel:
I forgot to test setting the config key to the non-default value.

Update: `snap set` rejects keys with `_` in their name :/

simondeziel:
@tomponline this time, setting the (new) key to its non-default value (`false`) worked and did the right thing.

snapcraft.yaml
```diff
@@ -31,6 +31,7 @@ description: |-
 
  Supported configuration options for the snap (snap set lxd [<key>=<value>...]):
 
+   - apparmor.unprivileged_restrictions_disable: Whether to disable restrictions on unprivileged user namespaces [default=true]
```
tomponline: I believe you are using hyphens now?

snapcraft.yaml
```diff
@@ -45,11 +45,11 @@ description: |-
    - lxcfs.loadavg: Start tracking per-container load average [default=false]
    - lxcfs.cfs: Consider CPU shares for CPU usage [default=false]
    - lxcfs.debug: Increase logging to debug level [default=false]
+   - minio.path: Path to the minio binary to use with LXD [default=""]
```
tomponline, line 49: @masnax in a separate PR please can you update description of this setting to make it clear its the path to the directory containing the minio and mc commands thanks

snapcraft.yaml
simondeziel: @tomponline I tested from the up to date repo but had not push. So test results are still good and I've since force pushed.

tomponline (approved):

PR #334, opened 2024-02-15
dqlite: Bump to v1.16.2 (latest-candidate)

tomponline:
Needed to work with SQLite versions greater than 3.45.0

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1383,9 +1383,10 @@ parts:
 
   lxd:
     source: https://github.com/canonical/lxd
-    source-depth: 1
-    source-type: git
+    # XXX: We often cherry-pick for candidate builds so don't do shallow clone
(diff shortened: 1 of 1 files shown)


tomponline:
@simondeziel this will fix the latest-candidate builds too

simondeziel (dismissed):

snapcraft.yaml
```diff
@@ -1383,7 +1383,6 @@ parts:
 
   lxd:
     source: https://github.com/canonical/lxd
-    source-depth: 1
```
simondeziel, line 1386: Mind putting that in comment with a quick explanation, just so that I don't undo this by mistake ;)

tomponline: done

simondeziel (approved):

PR #336, opened 2024-02-15
snapcraft: (re)prime usb.ids

simondeziel:
In commit 3b35370d53489d8ede56490522, only the pci.ids was primed to be explicit in what we needed but that missed the fact that usb.ids which comes from another package (deps of lshw) was also needed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1410,7 +1410,7 @@ parts:
       - pci.ids
       - pigz
       - rsync
-      - usbutils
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline I checked an old snap (5.0.2) to confirm that no bits of `usbutils` were primed before deciding to drop it. We only ever cared about `usb.ids` only.

tomponline (approved):

PR #339, opened 2024-02-19
Replace snap-query (Go) by python implementation

simondeziel:
This python implementation leverages the fact that python3 is shipped in the base core snap and avoids shipping the Go version of snap-query that was ~4.7MiB once compiled and stripped.


tomponline (commented):
What about using jq?

simondeziel:
Nah, `jq` isn't in the base snap, `python3` is though.

tomponline:
We could add it though? Would be smaller than the go program no?

tomponline:
Ah but Python is available I see

tomponline:
I would prefer correct over brittle tbh

simondeziel:
The go and python implementation were tested manually by comparing their output while doing:

* snap stop lxd
* snap start lxd
* snap restart lxd
* snap revert lxd
* snap refresh lxd --channel latest/stable
* systemctl stop snap.lxd.daemon

In all cases, the output was identical and what was logged by `daemon.stop` the same.

tomponline (approved):

PR #340, opened 2024-02-20
Only support ZFS 2.1 and up

simondeziel:
Ubuntu 22.04 is essentially the minimum we support for the next LXD LTS so since https://packages.ubuntu.com/jammy/zfsutils-linux ships version 2.1.5, we should drop the now unsupported zfs-0.8 and zfs-2.0 parts.


tomponline (approved):
Can you update the min requirements doc page with this too please.

snapcraft/commands/lxd
```diff
@@ -20,8 +20,8 @@ export PATH="${PATH}:${SNAP_CURRENT}/bin"
 export LXD_DIR="${LXD_DIR:-"${SNAP_COMMON}/lxd/"}"
 
 # Make sure we have a ZFS binary on the path
-export LD_LIBRARY_PATH="${SNAP_CURRENT}/zfs-0.8/lib/:${LD_LIBRARY_PATH}"
-export PATH="${SNAP_CURRENT}/zfs-0.8/bin:${PATH}"
+export LD_LIBRARY_PATH="${SNAP_CURRENT}/zfs-2.1/lib/:${LD_LIBRARY_PATH}"
+export PATH="${SNAP_CURRENT}/zfs-2.1/bin:${PATH}"
```
tomponline, line 24: what does this bit do btw? The default?

simondeziel: I believe so but I'm not 100% sure.

PR #342, opened 2024-02-20
Use latest ZFS tooling if no exact match is found

simondeziel:
This kind of logic would have avoided the rushed 5.0.2 respin to cope with Ubuntu 22.04 HWE kernel moving from 6.2.0 (ZFS 2.1) to 6.5.0 (ZFS 2.2).

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -351,7 +351,10 @@ if [ -d "${SNAP_CURRENT}/zfs-${ZFS_VER}/bin" ]; then
     export LD_LIBRARY_PATH="${SNAP_CURRENT}/zfs-${ZFS_VER}/lib/:${LD_LIBRARY_PATH}"
     export PATH="${SNAP_CURRENT}/zfs-${ZFS_VER}/bin:${PATH}"
 elif [ -n "${ZFS_VER}" ]; then
-    echo "==> Unsupported ZFS version (${ZFS_VER})"
+    NEWEST_ZFS_VER="$(cd "${SNAP_CURRENT}" && find -maxdepth 1 -type d -name 'zfs-*' | sort | tail -n1 | cut -d- -f2-)"
+    echo "==> Unsupported ZFS version (${ZFS_VER}), falling back to ${NEWEST_ZFS_VER}"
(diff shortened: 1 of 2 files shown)


simondeziel:
The fallback logic was test by forcing the exact version match branch to fail.

tomponline (approved):
Do you think there's a chance of data corruption with this change should lxd use an incompatible user land or is it just that some commands may fail?

simondeziel:
> Do you think there's a chance of data corruption with this change should lxd use an incompatible user land or is it just that some commands may fail?

As far as I know, the zfs/zpool commands are mostly used for configuring the pool so any incompat would likely result in a command execution error.

I can only think of `zfs send|recv` that could be more involved with data directly. I don't if they do anything more than just asking zfs.ko to produce or import ZFS data. I could try and ask them on IRC.

simondeziel:
@tomponline, I have mixed feelings about this:
```
18:20:08 	<sdeziel>	hello, we have a tool (LXD snap) that ships a bunch of zfs utils (user land bits only) and we use those to configure ZFS, send/recv FSes and such. We try to match the utils with the zfs.ko version but cannot always get an exact match. How risky is it to use 2.2 utils (zfs/zpool commands) to deal with a zfs.ko that is either older or newer?
18:21:37 	<sdeziel>	We are fine with `zfs set` or `zpool set` to give an error but would like to know if somehow user data could be at risk during say `zfs send|recv`.
18:33:12 	<PMT>	I don't know that there's any formal promises made about how well that will work, even in theory. It's intended to not break, but I don't really think anyone tests that very much, as demonstrated by 2.2's zpool binaries reporting incorrect zpool status on 2.1 kernel modules.
18:34:05 	<PMT>	(Or the strange assertion failures that happened when Ubuntu shipped a userland too old to understand compression=zstd with kernel modules that did.)
18:35:00 	<PMT>	So as far as I know, the answer is "it's supposed to work, and I think bugs reported with it not working will be treated as valid and fixed to the extent possible, but uh, I wouldn't try it if I could avoid it"
18:39:45 	<sdeziel>	PMT: thanks a bunch, that's valuable input
```

tomponline:
Thanks for checking. Doesn't fill me with confidence though.

PR #343, opened 2024-02-20
snapcraft: have dqlite build raft

simondeziel:
The dqlite folks have announced their support for building with `--enable-build-raft` allowing us to stop caring about raft directly.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -283,7 +283,6 @@ parts:
 
   dqlite:
     after:
-      - raft
       - sqlite
     source: https://github.com/canonical/dqlite
     source-depth: 1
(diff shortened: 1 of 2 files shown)


simondeziel:
This passed the `lxd-ci` cluster locally.

simondeziel:
This passed the `lxd-ci` cluster locally.

snapcraft.yaml
```diff
@@ -283,23 +283,27 @@ parts:
...
       - libuv1
     build-packages:
+      - liblz4-dev
       - libuv1-dev
     organize:
       usr/lib/: lib/
     prime:
       - lib/libdqlite*so*
       - lib/*/libuv*
+      - lib/libraft*so*
```
cole-miller: the dqlite build system doesn't build a separate libraft.so when `--enable-build-raft` is passed, everything is folded into libdqlite.so

simondeziel: @cole-miller many thanks, I just confirmed that indeed there is no such `libraft` lib.

tomponline (approved):
Thanks!

tomponline:
We cant move this over to latest-candidate until @cole-miller makes a dqlite release with this change.

tomponline:
@simondeziel please can you also update the `make deps` command and potentially this part of the workflows https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L25-L27  thanks

PR #344, opened 2024-02-21
Revert "snapcraft: have dqlite build raft"

simondeziel:
This reverts commit 3816410d75b6d4ab161d5fcf0123c9dfa5d35915.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -283,6 +283,7 @@ parts:
 
   dqlite:
     after:
+      - raft
       - sqlite
     source: https://github.com/canonical/dqlite
     source-depth: 1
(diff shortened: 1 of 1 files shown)


simondeziel:
@cole-miller just a FYI but we are reverting that unified build of dqlite+raft as we suspect it introduces a regression when running our OVN tests from [lxd-ci](https://github.com/canonical/lxd-ci):

```
# PURGE_LXD=1 ./bin/local-run tests/network-ovn latest/edge peering
...
==> Now setup peering
+ lxc network peer create ovn1 ovn1foo ovn2/ovn2 --project=ovn1
Error: Failed creating peer: sql: no rows in result set
```

tomponline:
@simondeziel lets land the change in our LXD test suite first before landing this the change in the snap again so we can check for regressions on the pull request tests.

tomponline (approved):

simondeziel:
@tomponline I'm afraid I don't know which bit of the LXD test suite you are referring to. If you are talking about `tests/network-ovn` from lxd-ci, then this part didn't change recently and it's in the early/first interactions with OVN so probably not a test issue.

tomponline:
I mentioned it in the other PR, but here by updating make deps to build the new combined dqlite
https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L108

and then potentially remove the raft references here
https://github.com/canonical/lxd/blob/main/.github/workflows/tests.yml#L25-L26

simondeziel:
@tomponline ah sorry, I didn't read right that you asked for those other changes to be done prior to introducing the unified dqlite+raft build **again**. Sure makes sense.

PR #352, opened 2024-02-29
Fallback to directly symlinking from MicroOVN

### LINKED ISSUE #13001: Ensure LXD snap detects all stable releases of MicroOVN
The commit https://github.com/canonical/lxd-pkg-snap/pull/275/commits/4bc76da44c6bbe2fe07c3b9912b646630d52bbf2 removed direct symlinking from MicroOVN in favour of using the content interfaces, but due to MicroOVN's release schedule, not every stable version of MicroOVN will necessarily support this at the same time as LXD. 

As such we need to ensure that we support both methods until such a time as MicroOVN supports the content interface on all its stable tracks.

masnax:
Closes https://github.com/canonical/lxd/issues/13001

Adds back direct symlinking of MicroOVN to LXD, as MicroOVN's release schedule means LXD will be updated before MicroOVN has the content interface in its default track. 

Simultaneously updates the content interface due to a race when specifying individual files in a content interface: https://github.com/canonical/microovn/pull/123

The new structure removes the old broken `ovn-api` content interface, and replaces it with the newer `ovn-certificates` content interface. 

The old interface will be ignored, so no error should happen. If the new interface doesn't exist, we will fall back to the old direct symlink method.


masnax:
@tomponline Should be ready for review now.

tomponline (approved):
Thanks

PR #380, opened 2024-03-15
Rename microovn content interface

masnax:
The microovn content interface for certificates is called `ovn-certificates` now, instead of the initially proposed `ovn-api`, which was chosen when it contained more than just certificates.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -65,7 +65,7 @@ plugs:
  ceph-conf:
    interface: content
    target: "$SNAP_DATA/microceph"
- ovn-api:
+ ovn-certificates:
    interface: content
    target: "$SNAP_DATA/microovn/certificates"
  ovn-chassis:


masnax:
For the record, since we have a fallback to direct symlinks in place from the last PR, this isn't necessarily breaking anything at the moment. But it is preventing 50% of the content interfaces that we want from being used.

tomponline (approved):
Thanks

Please can you open PRs for this on latest-candidate, 5.0-edge, 5.0-candidate, 5.21-edge, and 5.21-candidate

Hopefully this is the end of changes to this interface now.

tomponline:
Maybe we don't need the 5.0 branches done. I can't keep track of all the iterations we've been through on this to know what state it's in there.

masnax:
> Maybe we don't need the 5.0 branches done. I can't keep track of all the iterations we've been through on this to know what state it's in there.

Yeah, those don't have any of the content interface stuff in them at all.

PR #388, opened 2024-03-18
github: Update dependabot to check 5.21-edge

tomponline:


tomponline:
@simondeziel does this need to be in the 5.21-edge branch itself?

simondeziel (approved):

PR #414, opened 2024-04-15
snapcraft/hooks: Use mkdir -p for ovn interface

masnax:
The same hooks can run concurrently many times, so use `mkdir -p` to succeed if the path exists

### DIFF
--- a/snapcraft/hooks/connect-plug-ovn-certificates
+++ b/snapcraft/hooks/connect-plug-ovn-certificates
@@ -42,7 +42,7 @@ if [ -e "/etc/.lxd_generated" ]; then
   fi
 
   if ! [ "${ovn_builtin:-"false"}" = "true" ]; then
-    mkdir /etc/ovn
+    mkdir -p /etc/ovn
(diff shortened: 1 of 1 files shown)


simondeziel (approved):

tomponline:
Thanks @masnax I think I've seen this actually.

Do you know why it gets called concurrently?

tomponline (approved):

simondeziel:
> Do you know why it gets called concurrently?

Not sure about the why but here's the when during microcloud `service mismatch` test:

```
 + lxc exec micro02 -- sh -c snap install microceph microovn
2024-04-16T16:16:53Z INFO Waiting for conflicting change in progress: conflicting slot snap snapd, task "connect"
- Run hook connect-plug-ovn-certificates of snap "lxd" (run hook "connect-plug-ovn-certificates": mkdir: cannot create directory ‘/etc/ovn’: File exists)
```

masnax:
I think what's happening here is LXD is starting, so we are running the `daemon.start` hook which also sets up these paths, but at almost the same time we start MicroOVN so the `connect-plug-ovn-certificates` hook gets run at the same time.

tomponline:
makes sense thanks

PR #418, opened 2024-04-16
zfs: Reinstate ZFS 0.8 for Bionic and Focal support (5.21-edge)

tomponline:
Allows users tracking latest/stable on Bionic and Focal to switch to 5.21/stable and still use ZFS.

Bionic users will need to switch to the Bionic HWE kernel.

Tested using an `ubuntu-daily:bionic` VM.


simondeziel (approved):
LGTM

PR #422, opened 2024-04-18
snapcraft/commands/daemon.start: consistently use `mkdir -p /etc/ovn`

simondeziel:
MicroCloud tests often fail with:

```
 + lxc exec micro02 -- snap install microceph microovn
2024-04-18T02:29:04Z INFO Waiting for conflicting change in progress: conflicting slot snap snapd, task "connect"
- Run hook connect-plug-ovn-certificates of snap "lxd" (run hook "connect-plug-ovn-certificates": mkdir: cannot create directory ‘/etc/ovn’: File exists)
```

While the error presumably comes from the hook "connect-plug-ovn-certificates", that hook uses `mkdir -p` since f650f1ce0bae590244. The only other place where that dir is created by LXD is in `daemon.start` where for some reason, `-p` was not used contrary to most other dir creations elsewhere in that script.


tomponline (approved):

tomponline:
I've experienced this too. Thanks

PR #426, opened 2024-04-18
connect-plug-ovn-certificates: gracefully handle /etc/ovn being a symlink

simondeziel:


snapcraft/hooks/connect-plug-ovn-certificates
```diff
@@ -37,8 +37,10 @@ get_bool() {
 if [ -e "/etc/.lxd_generated" ]; then
   ovn_builtin=$(get_bool "$(snapctl get ovn.builtin)")
 
-  if test -d /etc/ovn ; then
```
simondeziel, line 40: `/etc/ovn` can be symlink pointing to `/var/lib/snapd/hostfs/etc/ovn`. Since `test -d`|`[ -d ... ]` follow symlinks, that `if` will succeed if the final directory `/var/lib/snapd/hostfs/etc/ovn` exists. In this case, is it really a good idea to do a `rm -rf` of it?

I think we should first check if it's a symlink `[ -L ... ]` (which does not follow links ;) ) and if not, only then check if it's a dir.

In other words, I'm asking if we should reverse the `if`/`elif` from that PR.

snapcraft/hooks/connect-plug-ovn-certificates
```diff
@@ -37,14 +37,17 @@ get_bool() {
...
+    echo "=> Removing /etc/ovn symlink"
     rm -f /etc/ovn
   elif [ -d /etc/ovn ]; then
+    echo "=> Removing /etc/ovn directory"
     rm -rf /etc/ovn
   fi
 
   if ! [ "${ovn_builtin:-"false"}" = "true" ]; then
+    echo "=> Configuring MicroOVN Content Interface"
     mkdir -p /etc/ovn
```
roosterfish, line 51: Do we still need the `-p` here? I am wondering if it would hide a bug in the cleanup if we still have a `/etc/ovn` directory at this stage?

tomponline (approved):

snapcraft/hooks/connect-plug-ovn-certificates
simondeziel: I think you are right, that `-p` is superfluous. However I'm on the fence on whether to drop it or not as ideally `daemon.start` should be doing the same, IMHO. I'm hesitant to touch `daemon.start` more :/

roosterfish: Ok fine for me, shouldn't be a blocker.

roosterfish (approved):

tomponline:
@simondeziel can you let me know when this is in latest/edge and tested working ta

simondeziel:
> @simondeziel can you let me know when this is in latest/edge and tested working ta

Looks like it passed the microcloud test suite this time!

PR #430, opened 2024-04-19
microovn fixes (latest-candidate)

tomponline:
https://github.com/canonical/lxd-pkg-snap/pull/426/


MusicDin (approved):

tomponline (commented):
@simondeziel i was just about to merge these and then it occurred to me:

Why are we deleting the symlink or internal ovn config dir on daemon.start?

Won't this mean that we will break any ongoing ovn configurations when LXD is refreshed? Sounds like a potential cause for some gnarly bugs to me.

PR #446, opened 2024-04-25
Rework how existing LXD PIDs are validated/discovered

simondeziel:
Change how LXD's `daemon.start` detects existing LXD processes left behind.

1. Optimize how `pgrep` is used to avoid needing to invoke it twice, once for regular daemons and another for debug ones.
2. Tweak `pgrep` invocation to only report PIDs that have an EUID of `0`. This ensures we don't accidentally get nested LXD PIDs because those should be in a different userns (for `unpriv`)
3. Once we have a candidate PID, make sure it's from a snap (`SNAP_NAME=lxd` line in `/proc/$pid/environ`)
4. Now that we have a probable LXD PID, make sure it is really a leftover process and not a nested LXD from a `privileged` container.

This last step involves writing short random data through the PID's mount namespace (`/proc/${pid}/root/var/snap/lxd/common/lxd/.validate`) and comparing it with our own view of the mount namespace by accessing the same file but directly (`/var/snap/lxd/common/lxd/.validate`).

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -628,9 +628,8 @@ for pid in $(pgrep -f "lxd(\.debug)? --logfile"); do
     # to a file using the `/proc` path and comparing the content seen from
     # the `/var` path, we can conclude if the LXD daemon is executing
     # in the same mount namespace as ours (safe to kill) or not (ignore).
-    rm -f "/var/snap/lxd/common/lxd/.validate"
-    touch "/proc/${pid}/root/var/snap/lxd/common/lxd/.validate" 2>/dev/null || true
-    if [ -e "/var/snap/lxd/common/lxd/.validate" ]; then
+    echo "$$" > "/proc/${pid}/root/var/snap/lxd/common/lxd/.validate" 2>/dev/null || true
+    if [ "$(cat /var/snap/lxd/common/lxd/.validate 2>/dev/null)" = "$$" ]; then
         echo "=> Killing conflicting LXD (pid=${pid})"
         kill -9 "${pid}" || true
         rm -f "/var/snap/lxd/common/lxd/.validate"
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -619,7 +619,7 @@ fi
 
 # LXD
 ## Check for existing LXDs
-for pid in $(pgrep -f "lxd --logfile" ; pgrep -f "lxd.debug --logfile"); do
+for pid in $(pgrep -f "lxd(\.debug)? --logfile"); do
     # Cheap confirmation that we likely have a LXD PID
     grep -q --line-regexp --fixed-strings --max-count=1 "SNAP_NAME=lxd" "/proc/${pid}/environ" || continue
 
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -619,7 +619,7 @@ fi
 
 # LXD
 ## Check for existing LXDs
-for pid in $(pgrep -f "lxd(\.debug)? --logfile"); do
+for pid in $(pgrep --euid 0 --full "lxd(\.debug)? --logfile"); do
     # Cheap confirmation that we likely have a LXD PID
     grep -q --line-regexp --fixed-strings --max-count=1 "SNAP_NAME=lxd" "/proc/${pid}/environ" || continue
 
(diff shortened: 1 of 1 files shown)


snapcraft/commands/daemon.start
```diff
@@ -620,7 +620,8 @@ fi
 # LXD
 ## Check for existing LXDs
 for pid in $(pgrep -f "lxd --logfile" ; pgrep -f "lxd.debug --logfile"); do
-    grep -qF "SNAP_NAME=lxd" "/proc/${pid}/environ" || continue
+    # Cheap confirmation that we likely have a LXD PID
+    grep -qxFm1 "SNAP_NAME=lxd" "/proc/${pid}/environ" || continue
```
tomponline: Are there long form arguments we can use here? Either that or an extension of the commit message/comment to explain how this is an improvement/cheaper?

simondeziel: While they all have long form versions, I think switching them all would make it harder to understand to some as most of those options I've only ever seen used in their short form.

I'm assuming the `-m1` bit is the one needing clarification here which I'm happy to expand as indeed it's not commonly used and the long form is quite clear (`--max-count=1`):
>        -m NUM, --max-count=NUM
>              Stop  reading  a  file  after  NUM matching lines.

tomponline: actually it was most of them :)

tomponline: but the use of --max-count makes sense to me now

snapcraft/commands/daemon.start
```diff
@@ -619,7 +619,7 @@ fi
 
 # LXD
 ## Check for existing LXDs
-for pid in $(pgrep -f "lxd(\.debug)? --logfile"); do
+for pid in $(pgrep --euid 0 --full "lxd(\.debug)? --logfile"); do
```
tomponline, line 622: in the example in your commit, why is server_pid 1293, but this command doesn't match that process?

tomponline: Also why not --uid?

snapcraft/commands/daemon.start
simondeziel: I've now expanded `-x` and `-F` as well.

snapcraft/commands/daemon.start
simondeziel: The `server_pid 1293` is relative to the container's PID NS which is why it doesn't match the host's view.

simondeziel: As to why not `--uid` instead of `--euid` it shouldn't really matter here as they should both be `0`. The distinction becomes relevant when using setuid binaries.

simondeziel: Also, in other scripts we use `id -u` to ensure the EUID is `0` so this should be more consistent.

snapcraft/commands/daemon.start
```diff
@@ -625,12 +625,11 @@ for pid in $(pgrep -f "lxd(\.debug)? --logfile"); do
 
     # Confirm the PID found by pgrep is indeed ours and not one from
     # a nested instance also running LXD or a recycled PID
-    rm -f "/var/snap/lxd/common/lxd/.validate"
-    touch "/proc/${pid}/root/var/snap/lxd/common/lxd/.validate" 2>/dev/null || true
-    if [ -e "/var/snap/lxd/common/lxd/.validate" ]; then
+    echo "$$" > "/proc/${pid}/root/var/snap/lxd/common/lxd/.validate" 2>/dev/null || true
```
tomponline, line 633: I'm not really following why this change is needed (leaving aside the microoptimisation for now too).

What issue is this fixing? Have we seen it occur? 

I'm confused why a nested LXD would be able to write to the /var/snap/lxd/common/lxd directory of its parent?

tomponline:
@simondeziel it'd be interesting to know the context of the precursor discussions you may have had with @mihalicyn about this (he's referenced in the commits) as im not really aware of the overall reason for this PR.

Adding a PR description would likely help to add context here.

snapcraft/commands/daemon.start
simondeziel: It's not fixing any known/observed issue more a theoretical/potential one.

This came about after I asked Aleks if it was OK to replace the touch+rm by just checking for `"/proc/${pid}/root/var/snap/lxd/common/lxd/server.key`'s presence.

Aleks quickly pointed out that would defeat the purpose of that check which is to make sure we are dealing with a LXD PID that belongs to our PID and mount NS. That's because any nested LXD PID we'd get from that `pgrep` would also lead us to a `server.key` but in a different mount NS. Hence the writing to it was needed.

I then observed that if the nested LXD was also going through that same validation at the same time, it could also race and create the `.validate` file to causing problems.

To which Aleks replied: "write and compare random data to eliminate the race".

simondeziel:
@tomponline I've updated the comment based on our discussion to better explain that the `.validate` dance is to rule out *priv* nested LXD by making sure the PID in question doesn't use the same mount namespace as "ours".

I've also reintroduced the removal of that `.validate` file but differently to avoid "leaking" some of them on inside directories of "failed" candidate LXD.

snapcraft/commands/daemon.start
simondeziel: The micro-optimization was dropped.

tomponline (approved):

PR #454, opened 2024-05-06
Disable some uneeded QEMU config flags

simondeziel:


snapcraft.yaml
```diff
@@ -815,6 +815,7 @@ parts:
       - --disable-parallels
       - --disable-qed
       - --disable-slirp
+      - --disable-tcg
```
tomponline: I understand this is needed because we use qemu to setup uefi keys during snap build. Which doesn't have kvm.available.

snapcraft.yaml
```diff
@@ -820,6 +820,7 @@ parts:
       - --disable-vdi
       - --disable-vnc
       - --disable-xen
+      - --disable-xkbcommon
```
tomponline, line 823: What does this do, considerations of impact?

snapcraft.yaml
```diff
@@ -813,6 +813,7 @@ parts:
       - --disable-docs
       - --disable-guest-agent
       - --disable-parallels
+      - --disable-pvrdma
```
tomponline, line 816: We need to understand impact and what this does.

tomponline (changes requested):
Commit messages could do with being expanded a bit.

snapcraft.yaml
simondeziel: Ah, that could explain why `qemu-ovmf-secureboot` no longer builds with this, too bad!

snapcraft.yaml
simondeziel: The markdown formatting eat the relevant bit of the commit, I've put them back now.

snapcraft.yaml
simondeziel: The markdown formatting ate the relevant bit of the commit, I've put them back now.

snapcraft.yaml
tomponline: That's the one. @mihalicyn has an item for this but depends on secure boot keys being available as a package in ubuntu.

snapcraft.yaml
```diff
@@ -819,6 +819,7 @@ parts:
       - --disable-vdi
       - --disable-vnc
       - --disable-xen
+      - --disable-xkbcommon
```
tomponline, line 823: Do we also need the key map files too?

tomponline (approved):

snapcraft.yaml
simondeziel: I don't know if the keymap files are needed/used by anything. Maybe @mihalicyn would know?

PR #456, opened 2024-05-11
Use variable substitution to replace `basename` that is an external binary

simondeziel:
`basename` is not a builtin command so replacing it with variable substitution is a bit less readable but avoids forking.


tomponline (approved):
Thanks for the comments.

PR #470, opened 2024-06-20
snapcraft: Add `nvidia-ctk` as part of the binary tool for Container Device Interface spec generation

gabrielmougard:
related to: https://github.com/canonical/lxd/pull/13562

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -650,6 +650,32 @@ parts:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container*.so*
 
+  nvidia-container-toolkit:
+    source: https://github.com/NVIDIA/nvidia-container-toolkit
+    source-depth: 1
+    source-commit: a470818ba7d9166be282cd0039dd2fc9b0a34d73 # v1.16.1
+    source-type: git
+    build-snaps:
+      - go
+    plugin: make
+    override-prime: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      craftctl default
+    override-pull: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      craftctl default
+    override-build: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      set -ex
+
+      make binaries
+      mkdir -p "${CRAFT_PART_INSTALL}/bin/"
+      cp nvidia-ctk "${CRAFT_PART_INSTALL}/bin/"
+    organize:
+      usr/bin/: bin/
+    prime:
+      - bin/nvidia-ctk
+
   nvme:
     plugin: nil
     stage-packages:
@@ -1479,6 +1505,7 @@ parts:
       - criu
       - lxd
       - shmounts
+      - nvidia-container-toolkit
     plugin: nil
     override-prime: |
       set -x


snapcraft.yaml
```diff
@@ -650,6 +650,30 @@ parts:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container*.so*
 
+  nvidia-container-toolkit:
```
tomponline, line 653: this is missing the compiler tool to use, is it Go? Take a look at the lxd section for examples.

gabrielmougard: This is weird because I built the snap like this and it did not cause any issue even without a `build-packages: go`.. And yes the compiler to use is Go.

tomponline: Its probably relying on an earlier part that installed Go - which works but is brittle.

tomponline: To be clearer, we still need to add the `build-packages: go` here too please

simondeziel: I agree that we should explicitly depend on Go, however I think we want:

```
    build-snaps:
      - go
```

snapcraft.yaml
```diff
@@ -650,6 +650,30 @@ parts:
...
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
+      set -ex
+
+      make binaries
+      mkdir -p "${CRAFT_PART_INSTALL}/bin/"
+      cp nvidia-ctk "${CRAFT_PART_INSTALL}/bin/"
+    organize:
+      usr/bin/: bin/
+    prime:
+      - bin/nvidia-ctk
```
simondeziel, line 677: Is this a static binary (as in 0 `.so` files showing in `ldd`'s output)?

simondeziel (changes requested):
Aside from the `build-snaps` directive that's missing, I think this part needs to be added to the `after` list of the `strip` part.

snapcraft.yaml
gabrielmougard: no it's not:
```
$ ldd nvidia-ctk 
        linux-vdso.so.1 (0x00007ffcaa0d0000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000072888c400000)
        /lib64/ld-linux-x86-64.so.2 (0x000072888c6c5000)
```

snapcraft.yaml
gabrielmougard: fixed

gabrielmougard:
> Aside from the `build-snaps` directive that's missing, I think this part needs to be added to the `after` list of the `strip` part.

done :)

snapcraft.yaml
simondeziel: Ah thanks for checking. I think those are special `.so` you don't need to prime so I think this is all good then.

simondeziel (dismissed):

snapcraft.yaml
gabrielmougard: fixed

snapcraft.yaml
```diff
@@ -650,6 +650,32 @@ parts:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container*.so*
 
+  nvidia-container-toolkit:
+    source: https://github.com/NVIDIA/nvidia-container-toolkit
+    source-depth: 1
+    source-tag: v1.15.0
+    source-type: git
+    build-snaps:
+      - go
```
tomponline, line 659: Please use go/1.22/stable

tomponline: oh ignore me, sorry, we only pin on candidate builds.

snapcraft.yaml
```diff
@@ -650,6 +650,32 @@ parts:
...
+    plugin: make
+    override-prime: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
```
tomponline: Does this only work on amd64 and arm64 then?

snapcraft.yaml
```diff
@@ -650,6 +650,32 @@ parts:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container*.so*
 
+  nvidia-container-toolkit:
+    source: https://github.com/NVIDIA/nvidia-container-toolkit
+    source-depth: 1
+    source-tag: v1.15.0
```
tomponline: Please can you use 

```
    source-commit: <commit> # v<version>
```

style.

snapcraft.yaml
gabrielmougard: Actually, it seems to also be supported for `ppc64le`

gabrielmougard: ... and `s390x` (it was not clearly stated in the repo but after a bit of digging I found this: https://rpmfind.net/linux/rpm2html/search.php?query=nvidia-container-toolkit)

tomponline: its just Go right?

tomponline: I'm fine with limiting it if you think there is a reason that this tool wouldn't be used on those architectures, even if it can be built. Can iGPUs be used on ppc64le and s390x?

snapcraft.yaml

snapcraft.yaml
gabrielmougard: No idea.. I'm asking.

gabrielmougard: > its just Go right?

No, there is a small C lib here: https://github.com/NVIDIA/nvidia-container-toolkit/blob/a470818ba7d9166be282cd0039dd2fc9b0a34d73/internal/dxcore/dxcore.c#L7

gabrielmougard: I asked @elezar

snapcraft.yaml
```diff
@@ -650,6 +650,32 @@ parts:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container*.so*
 
+  nvidia-container-toolkit:
+    source: https://github.com/NVIDIA/nvidia-container-toolkit
+    source-depth: 1
+    source-commit: a470818 # v1.16.1
```
tomponline: full commit hash please

tomponline (approved):

snapcraft.yaml
elezar: I'm OOTO. Will get to this on Tuesday.

elezar: It is my understanding that iGPUs are `aarch64` only. For discrete GPUs our main platforms are `x86_64` and `aarch64`. We do also build `ppc64le` packages, but it is my understanding that driver support for these architectures is being deprecated https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#deprecated-or-dropped-architectures

Note that this does not mean that CDI should not be supported on `ppc64le` and `s360x` platforms, but just that the tooling to generate NVIDIA CDI specifications -- which relies on the driver -- may not work there if the driver is not available.  One should still be able to build the go binary on these platforms though as device and library discovery using the driver is a runtime concern.

elezar: Note that the linked c file should not be relevant on these platforms either since these are specific to discovering the location of driver libraries on WSL2-based systems. If the dxcore.so shared libraries are not present, these are never used.

snapcraft.yaml
```diff
@@ -650,6 +650,32 @@ parts:
...
+    override-pull: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      craftctl default
+    override-build: |-
+      [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && exit 0
+      set -ex
+
+      make binaries
+      mkdir -p "${CRAFT_PART_INSTALL}/bin/"
+      cp nvidia-ctk "${CRAFT_PART_INSTALL}/bin/"
```
elezar, line 673: Note: If you're only interested in the `nvidia-ctk` binary (or the `nvidia-cdi-hook`, I would assume), you should be able to run `make cmd-nvidia-ctk` or `make cmd-nvidia-cdi-hook` instead of `make binaries` to only build the single binary.

tomponline: @gabrielmougard what do you think?

thanks @elezar !

snapcraft.yaml
tomponline: Thanks @elezar 

@gabrielmougard are you able to address this please?

PR #471, opened 2024-06-20
snapcraft: switch to core24

mihalicyn:
Tested locally. I was able to run container & VM without any problems.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -621,6 +621,7 @@ parts:
       - libelf-dev
       - libseccomp-dev
       - lsb-release
+      - libtirpc-dev
     override-prime: |-
       [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && exit 0
       craftctl default
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -553,7 +553,7 @@ parts:
   nasm:
     source: https://github.com/netwide-assembler/nasm
     source-depth: 1
-    source-tag: nasm-2.16.01
+    source-tag: nasm-2.16.03
     source-type: git
     plugin: autotools
     autotools-configure-parameters:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -662,7 +662,7 @@ parts:
   openvswitch:
     source: https://github.com/openvswitch/ovs
     source-depth: 1
-    source-tag: v3.3.0
+    source-tag: v3.3.1
     source-type: git
     plugin: autotools
     autotools-configure-parameters:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1120,7 +1120,7 @@ parts:
   zfs-2-2:
     source: https://github.com/openzfs/zfs
     source-depth: 1
-    source-tag: zfs-2.2.3
+    source-tag: zfs-2.2.4
     source-type: git
     plugin: autotools
     autotools-configure-parameters:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: lxd
-base: core22
+base: core24
 assumes:
   - snapd2.39
 version: git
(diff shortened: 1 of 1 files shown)


tomponline (commented):
Please can you separate the dependency bumps in to separate commits from the core24 bump ta

mihalicyn:
Have done.

snapcraft.yaml
```diff
@@ -1120,7 +1120,7 @@ parts:
   zfs-2-2:
     source: https://github.com/openzfs/zfs
     source-depth: 1
-    source-tag: zfs-2.2.3
+    source-tag: zfs-2.2.4
```
markylaing, line 1124: We discussed in the stand up that it would be good to perform more verification on these versions, e.g.:
- Are there any outstanding security issues?
- Can we make snapcraft verify a hash/GPG after downloading it?

MusicDin: It seems that snapcraft supports `source-checksum` which accepts `algorithm/digest`.

https://snapcraft.io/docs/snapcraft-parts-metadata#heading--source

I'm not sure if this works with git repo, but maybe we can use the checksum of the source code archive?

tomponline (changes requested):
Please can each dependency bump be its own commit like how its normally done.
E.g. https://github.com/canonical/lxd-pkg-snap/commit/c98308d85e6b57595836317316f72cd522201687

snapcraft.yaml
simondeziel: Thanks for bringing that up it makes sense. This `source-checksum` seems like the way to go, could be saved for another PR too.

tomponline: yes this isnt relevant for this PR - also a question I have is which hash is used for the source hash?
The source is downloaded over HTTPS using a specific git tag, so not clear which hash to use.
The release checksum will be presumably the checksum of the tarball not the git repo source.

simondeziel: Good point re the checksum being for the tarball. When we take care of this in another PR, `source-commit` could be used to reference the full SHA1 hash of the commit associated with the tag. In fact we might want to remove `source-tag` altogether and use:

```yaml
  zfs-2-2:
    source: https://github.com/openzfs/zfs
    source-depth: 1
    source-commit: 2566592045780e7be7afc899c2496b1ae3af4f4d # zfs-2.2.4
```

tomponline: That seems confusing to me - why not let the publisher decide which commit matches to a version?

simondeziel: While I agree that's less readable, it protects us from any modifications of the tag. GitHub recommends doing so for 3rd party actions for example: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

https://michaelheap.com/ensure-github-actions-pinned-sha/ goes into more details about this ^.

tomponline: OK makes sense. Thanks for the links! but yes lets do this for all deps in a separate PR and not conflate them.

simondeziel: Good. I'll also put up a Jira card to do the same with our various GH Actions.

snapcraft.yaml
```diff
@@ -1014,6 +1014,7 @@ parts:
     source-type: git
     plugin: rust
     build-packages:
+      - libtirpc-dev
```
simondeziel: Presumably that dependency was implicit when we used `core22`, right? If so, having this in a separated commit would make it easier to backport to other branches. Thanks

mihalicyn: Thanks! Fixed.

tomponline (approved):
Thanks!

snapcraft.yaml
MusicDin: > Good. I'll also put up a Jira card to do the same with our various GH Actions.

May be worth investigating Renovate for GH actions (which we already use in quite few repositories). Renovate allows pinning the GH action version - it will update commit hash of the GH action and add a comment next to it containing the actual version the hash refers to.

It is one of these two options:
- https://docs.renovatebot.com/presets-helpers/#helperspingithubactiondigeststosemver
- https://docs.renovatebot.com/presets-helpers/#helperspingithubactiondigests

PR #474, opened 2024-06-21
[WIP] snapcraft: add apparmor part for core22 only (5.0-edge)

mihalicyn:
Let's ship newer version of AppArmor with support of new features. Only needed for core22.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1209,15 +1209,50 @@ parts:
       - bin/pzstd
       - bin/zstd
 
+  apparmor:
+    source: https://gitlab.com/apparmor/apparmor.git
+    source-depth: 1
+    source-tag: v4.0.1
+    source-type: git
+    plugin: autotools
+    build-packages:
+      - g++
+      - bison
+      - flex
+      - autoconf-archive
+    override-build: |-
+      set -ex
+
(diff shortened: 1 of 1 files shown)


tomponline (commented):
We shouldn't need this for latest edge as you just switched that to core24. Instead this pr should target the 5.0 edge branch which is being transitioned to core22 by @simondeziel

mihalicyn:
yes, of course. I just decided to keep it there so we can discuss it. Also, I had to add some support to LXD to choose this LXD-snap-shipped version instead of the default provided in core22. So, it is a work in progress for now.

tomponline:
> yes, of course. I just decided to keep it there so we can discuss it. Also, I had to add some support to LXD to choose this LXD-snap-shipped version instead of the default provided in core22. So, it is a work in progress for now.

OK I'll wait for your update, although I am concerned about having to modify LXD, can we avoid that by doing changes in the snap to set the PATH to the apparmor parser? 

Anyway I'll wait until I can see what you're getting at :)

mihalicyn:
> OK I'll wait for your update, although I am concerned about having to modify LXD, can we avoid that by doing changes in the snap to set the PATH to the apparmor parser?

you are absolutely right. It just works with current `PATH` variable value ;-)

PR #475, opened 2024-06-21
snapcraft: install edk2 build dependencies only when needed

mihalicyn:
It's a fix for a build failure on s390:
Installing build-packages
Cannot find package listed in 'build-packages': acpica-tools Full execution log: '/root/.local/state/snapcraft/log/snapcraft-20240621-073250.087743.log' Build failed

Reason is that, in Ubuntu 24.04 acpica-tools package is not available for s390: https://packages.ubuntu.com/noble/acpica-tools
compare with:
https://packages.ubuntu.com/jammy/acpica-tools

Let's workaround this by preventing build packages installation for any architectures other than amd64/arm64. It's safe, because we only build edk2 for these two architectures.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -303,9 +303,10 @@ parts:
     source-type: git
     plugin: nil
     build-packages:
-      - g++
-      - acpica-tools
-      - uuid-dev
+      - on amd64,arm64:
+        - g++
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks!

PR #479, opened 2024-06-21
snapcraft: drop ceph support from `armhf`

simondeziel:
With the move to `core24`, the ceph package is no longer available on 32 bit arches. The intersection of users on `armhf` and using ceph storage pool is deemed to be exceedingly small anyway.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -191,10 +191,26 @@ parts:
       - libatomic
     plugin: nil
     stage-packages:
-      - ceph-common
+      - on amd64:
+        - ceph-common
+      - on arm64:
+        - ceph-common
+      - on ppc64el:
+        - ceph-common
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -195,6 +195,15 @@ parts:
     organize:
       usr/bin/: bin/
       usr/lib/: lib/
+    override-prime: |-
```
mihalicyn, line 205: Does this skips installation of `stage-packages`? As far as I understand `stage-packages` will be tried to install anyways :(

mihalicyn (dismissed):

snapcraft.yaml
simondeziel: Thanks, fixed now

tomponline (approved):

PR #486, opened 2024-06-26
Prepare 4.0.10 bis (4.0-edge)

simondeziel:


snapcraft.yaml
```diff
@@ -1048,7 +1048,7 @@ parts:
       - libseccomp
     source: https://github.com/lxc/lxc
     source-type: git
-    source-branch: stable-4.0
+    source-tag: lxc-4.0.12
```
tomponline, line 1051: Should we use hashes ala the new policy?

tomponline (approved):

snapcraft.yaml
simondeziel: I wanted to do it from newest to oldest (`latest-edge` -> `latest-candidate` -> ... -> `5.0-candidate`) and assumed 4.0 not going to see anymore releases after the upcoming 4.0.10 version.

It's probably a safe change to make so tell me if you want it now/here.

simondeziel: I just merged it as-is so that I can progress with testing. Happy to add another PR to make the switch on that `4.0-edge` branch.

tomponline: I dont mind, it might be useful to see the specific hashes of all deps though

PR #505, opened 2024-07-11
external QEMU snap support

mihalicyn:
requires https://github.com/canonical/lxd/pull/13742

https://warthogs.atlassian.net/browse/LXD-1164

### DIFF
--- a/lxd-qemu-snap/README.md
+++ b/lxd-qemu-snap/README.md
@@ -1,0 +1,69 @@
+# External QEMU snap for LXD snap
+
+## How to build and use
+
+```
+cd lxd-qemu-snap
+
+# clean up previous builds
+rm -f qemu-for-lxd_*.snap
+
+# build
+snapcraft
+
+# install
+sudo snap install qemu-for-lxd_*.snap --devmode
+
+# connect snaps
+sudo snap connect lxd:gpu-2404 mesa-2404:gpu-2404
+sudo snap connect lxd:qemu-external qemu-for-lxd:qemu-external
+```
+
+## How to use with virgl (only specific to this example of snapcraft.yaml)
+
+```
+lxc init images:ubuntu/noble/desktop desktop -c limits.memory=8GiB --vm
+
+# modify instance configuration:
+lxc config edit desktop
+
+# choose an appropriate renderer:
+ls -la /dev/dri/by-path/*-render
+
+# for example, on my system it is /dev/dri/by-path/pci-0000:67:00.0-render
+# lspci | grep -E "(3D|VGA)" shows:
+# 67:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Rembrandt (rev d8)
+
+# add the following lines:
+  raw.apparmor: |-
+    /snap/lxd/*/gpu-2404/** mr,
+    /dev/dri/ r,
+    /dev/dri/card[0-9]* rw,
+    /dev/dri/renderD[0-9]* rw,
+    /run/udev/data/c226:[0-9]* r,  # 226 drm
+    /sys/devices/** r,
+    /sys/bus/** r,
+  raw.qemu: -display egl-headless,rendernode=/dev/dri/by-path/pci-0000:67:00.0-render
+  raw.qemu.conf: |-
+    [device "qemu_gpu"]
+    driver = "virtio-vga-gl"
+
+# try it
+lxc start desktop --console=vga
+
+# you can check output from:
+dmesg | grep -i drm
+# if you see:
+# [drm] features: +virgl ...
+# it's a good sign
+
+glxinfo | grep -i vir
+# it should show something like:
+# Device: virgl ...
+```
+
+## References:
+
+https://github.com/snapcore/snapd/blob/5c8d8431baa425464b279ff26b8c44eecb9aab22/interfaces/builtin/opengl.go#L41
+
+https://gitlab.gnome.org/GNOME/gnome-boxes/-/issues/586
--- a/lxd-qemu-snap/snapcraft.yaml
+++ b/lxd-qemu-snap/snapcraft.yaml
@@ -1,0 +1,235 @@
+name: qemu-for-lxd
+base: core24
+assumes:
+  - snapd2.63
+version: '1'
+grade: devel
+summary: External Qemu for LXD
+description: |-
+ External Qemu for LXD
+
+contact: lxd@lists.canonical.com
+issues: https://github.com/canonical/lxd/issues
+source-code: https://github.com/canonical/lxd
+website: https://ubuntu.com/lxd
+confinement: strict
+
+slots:
+  qemu-external:
+    interface: content
+    content: qemu-external-binaries
+    read:
+      - $SNAP/external/qemu
+
+parts:
+  #
+  # This step is extremely important, as it ensures that no mesa libs will
+  # be provided with this snap to prevent any possible version mistmatches
+  # between this snap and mesa-2404 snap.
+  # If you the error like:
+  # DRI driver not from this Mesa build ('24.0.5-1ubuntu1' vs '24.0.9-0ubuntu0.1')
+  # this is a clear sign that you likely ship some mesa library along with your qemu-for-lxd
+  # snap and it has a version conflict with what you have in gpu-2404 interface snap.
+  #
+  gpu-2404:
+    after:
+      - virgl
+      - qemu
+    source: https://github.com/canonical/gpu-snap.git
+    plugin: dump
+    override-prime: |
+      set -ex
+
+      CUSTOM_PREFIX="external/qemu"
+      CRAFT_PRIME_OVERRIDE="${CRAFT_PRIME}/${CUSTOM_PREFIX}"
+
+      # cleanup script expects to see libraries in usr/lib, while we have them in lib
+      mkdir "${CRAFT_PRIME_OVERRIDE}/usr"
+      ln -s "${CRAFT_PRIME_OVERRIDE}/lib" "${CRAFT_PRIME_OVERRIDE}/usr/lib"
+
+      # a bit hacky way to call a cleanup script forcing it to look in a right directory
+      CRAFT_PRIME="${CRAFT_PRIME_OVERRIDE}" ${CRAFT_PART_SRC}/bin/gpu-2404-cleanup mesa-2404 nvidia-2404
+
+      rm -rf "${CRAFT_PRIME_OVERRIDE}/usr"
+
+      set +ex
+
+  virgl:
+    source: https://gitlab.freedesktop.org/virgl/virglrenderer.git
+    source-tag: virglrenderer-1.0.1
+    source-depth: 1
+    plugin: meson
+    meson-parameters:
+      - --prefix=/external/qemu
+    build-packages:
+      - meson
+      - libgbm-dev
+      - libdrm-dev
+
+  qemu:
+    after:
+      - virgl
+    source: https://gitlab.com/qemu-project/qemu.git
+    source-tag: v9.0.1
+    source-depth: 1
+    plugin: autotools
+    autotools-configure-parameters:
+      - --prefix=/external/qemu
+      - --enable-opengl
+      - --enable-virglrenderer
+      - --disable-smartcard
+# options from LXD snap:
+      - --disable-bochs
+      - --disable-cloop
+      - --disable-dmg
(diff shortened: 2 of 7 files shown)


lxd-qemu-snap/README.md
```diff
@@ -0,0 +1,72 @@
...
+# for example, on my system it is /dev/dri/by-path/pci-0000:67:00.0-render
+# lspci | grep 67:00 shows:
+# 67:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Rembrandt (rev d8)
+
+# add the following lines:
+  raw.apparmor: |-
+    /snap/lxd/*/gpu-2404/** mr,
+    /dev/dri/{,**} rw,
+    /dev/dri/card[0-9]* rw,
+    /dev/dri/renderD[0-9]* rw,
```
simondeziel, line 42: Aren't those 2 rules already covered by the `/dev/dri/{,**} rw,` one before?

lxd-qemu-snap/README.md
```diff
@@ -0,0 +1,72 @@
...
+lxc init images:ubuntu/noble/desktop desktop -c limits.memory=8GiB --vm
+
+# modify instance configuration:
+lxc config edit desktop
+
+# choose an appropriate renderer:
+ls -la /dev/dri/by-path/*-render
+
+# for example, on my system it is /dev/dri/by-path/pci-0000:67:00.0-render
+# lspci | grep 67:00 shows:
```
simondeziel:
```suggestion
# lspci | grep VGA shows:
```

snapcraft/commands/daemon.start
```diff
@@ -180,6 +180,12 @@ echo "==> Preparing /run/bin"
 mkdir -p "/run/bin"
 export PATH="/run/bin:${PATH}"
 
+if [ "${qemu_external_snap:-"false"}" = "true" ]; then
+    echo "==> Setting up external QEMU snap integration"
+    export LD_LIBRARY_PATH="${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}:${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}/pulseaudio:${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}/ceph:${LD_LIBRARY_PATH}"
```
simondeziel: I see pulseaudio in there and I wonder about pipewire?

mihalicyn: it's because QEMU was failing on attempt to load necessary libs from `${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}/pulseaudio`. Do you think we should add `${SNAP_CURRENT}/opt/external-qemu/lib/x86_64-linux-gnu/pipewire-0.3/` too?

lxd-qemu-snap/README.md
mihalicyn: it's not a mistake, actually, I wanted to show what's behind `67:00` from `ls -la /dev/dri/by-path/*-render`.

Let's do it even better:
`lspci | grep -E "(3D|VGA)"`

What do you think?

lxd-qemu-snap/README.md
mihalicyn: sure, let's do even better:
```
    /snap/lxd/*/gpu-2404/** mr,
    /dev/dri/ r,
    /dev/dri/card[0-9]* rw,
    /dev/dri/renderD[0-9]* rw,
    /run/udev/data/c226:[0-9]* r,  # 226 drm
    /sys/devices/** r,
    /sys/bus/** r,
```

snapcraft/commands/daemon.start
simondeziel: I don't know pipewire at all, but maybe? If so, could we use a shell glob to avoid hardcoding the version in the dir name?

lxd-qemu-snap/README.md
simondeziel: Oh, I see, I hadn't read the previous line properly. But sure, that `grep -E` LGTM.

lxd-qemu-snap/README.md
```diff
@@ -0,0 +1,72 @@
+# External QEMU snap for LXD snap
+
+## How to build and use
+
+```
+cd lxd-qemu-snap
+
+# clean up previous builds
+rm qemu-for-lxd_*.snap
```
simondeziel:
```suggestion
rm -f qemu-for-lxd_*.snap
```
For those blindly copy-n-pasting code snippets ;)

simondeziel (commented):
LGTM in general even if I don't like that `/opt` I won't object to it ;)

snapcraft/commands/daemon.start
mihalicyn: >I don't know pipewire at all, but maybe?

Unfortunately, I don't know too ;-)

>If so, could we use a shell glob to avoid hardcoding the version in the dir name?

Hm, something like:
```
LD_LPATH_PIPEWIRE="$(readlink -f "${SNAP_CURRENT}"/opt/external-qemu/lib/"${ARCH}"/pipewire-*/)"
export LD_LIBRARY_PATH="${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}:${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}/pulseaudio:${SNAP_CURRENT}/opt/external-qemu/lib/${ARCH}/ceph:${LD_LPATH_PIPEWIRE:+${LD_LPATH_PIPEWIRE}:}${LD_LIBRARY_PATH}"
```

?

simondeziel: Yeah, that `readlink` works.

lxd-qemu-snap/README.md
mihalicyn: Fixed. Thanks!

snapcraft/commands/daemon.start
mihalicyn: Have fixed!

mihalicyn:
Reworked to use connect/disconnect hooks.

simondeziel (dismissed):
LGTM, thanks!

mihalicyn:
Have added reload logic to connect/disconnect hooks.

lxd-qemu-snap/snapcraft.yaml
```diff
@@ -0,0 +1,213 @@
...
+
+slots:
+  external-qemu:
+    interface: content
+    content: external-qemu-binaries
+    read:
+      - $SNAP/opt/external-qemu
+
+parts:
+  gpu-2404:
```
tomponline, line 34: Is this section needed? What is it used for?

mihalicyn: It does a cleanup of libraries on the prime step in snap to prevent collisions with what `gpu-2404` snap provides.

snapcraft.yaml
```diff
@@ -1500,6 +1516,18 @@ parts:
     prime:
       - bin/setup-shmounts
 
+  gpu-2404:
```
tomponline, line 1519: Does this not get provided by the connected gpu-2404 snap? Whyy do we need this pulled here too?

mihalicyn: no, this is not provided by `gpu-2404`. This repo (https://github.com/canonical/gpu-snap) contains tools to help integration with `gpu-2404` interface snaps. This step, specifically, needed to clean up libraries on the prime step to prevent duplicates.

mihalicyn: You can read about it in the official guide https://mir-server.io/docs/the-gpu-2404-snap-interface

lxd-qemu-snap/snapcraft.yaml
```diff
@@ -0,0 +1,213 @@
...
+    after:
+      - qemu
+    source: https://github.com/canonical/gpu-snap.git
+    plugin: dump
+    override-prime: |
+      craftctl default
+      ${CRAFT_PART_SRC}/bin/gpu-2404-cleanup mesa-2404 nvidia-2404
+    prime:
+      - bin/gpu-2404-wrapper
```
simondeziel: Should the `-cleanup` script also be primed?
The [gpu-2404-cleanup](https://github.com/canonical/gpu-snap/blob/main/bin/gpu-2404-cleanup) script is a symlink to [graphics-core22-cleanup](https://github.com/canonical/gpu-snap/blob/main/bin/graphics-core22-cleanup). I don't know if the `prime` logic is smart enough to cope with that.

snapcraft.yaml
```diff
@@ -1500,6 +1516,18 @@ parts:
...
+    after:
+      - lxd
+      - qemu
+    source: https://github.com/canonical/gpu-snap.git
+    plugin: dump
+    override-prime: |
+      craftctl default
+      ${CRAFT_PART_SRC}/bin/gpu-2404-cleanup mesa-2404 nvidia-2404
+    prime:
+      - bin/gpu-2404-wrapper
```
simondeziel: Is the `-cleanup` needed here?

lxd-qemu-snap/snapcraft.yaml
mihalicyn: I don't think that we need to prime it as we just want to use it once during snap build time. And we doing it in the line `${CRAFT_PART_SRC}/bin/gpu-2404-cleanup mesa-2404 nvidia-2404`

simondeziel: OK, I hadn't realize it was a copy-n-paste from https://mir-server.io/docs/the-gpu-2404-snap-interface
Have you been able to test this? Cause when reading the `gpu-2404-cleanup` script, I think it will error out due to `nvidia-24.04` not existing in the [lists directory](https://github.com/canonical/gpu-snap/tree/main/lists).

mihalicyn: nope, I'm not specifically tested how this script works. I haven't seen any build time errors or warnings.

mihalicyn: I believe that `nvidia-2404` will be added soon and as we pull it in dynamically (without tag/version specifier) it will work properly.

simondeziel (dismissed):

snapcraft.yaml
```diff
@@ -75,6 +75,10 @@ plugs:
    interface: content
    target: $SNAP/gpu-2404
 #   default-provider: mesa-2404
+ external-qemu:
```
tomponline: Looking at the other plugs, and the snap feature options on https://snapcraft.io/lxd it seems like we usually use `noun.adjective` format, so this would be more in keeping to be called `qemu-external` in my view.

tomponline (changes requested):
Apart from the nit around naming consistency this looks good, thanks!

snapcraft.yaml
mihalicyn: Thanks! Have changed that.

lxd-qemu-snap/snapcraft.yaml
```diff
@@ -0,0 +1,235 @@
...
+    plugin: dump
+    override-prime: |
+      set -ex
+
+      CUSTOM_PREFIX="external/qemu"
+      CRAFT_PRIME_OVERRIDE="${CRAFT_PRIME}/${CUSTOM_PREFIX}"
+
+      # cleanup script expects to see libraries in usr/lib, while we have them in lib
+      mkdir "${CRAFT_PRIME_OVERRIDE}/usr"
+      ln -s "${CRAFT_PRIME_OVERRIDE}/lib" "${CRAFT_PRIME_OVERRIDE}/usr/lib"
```
simondeziel, line 48: Would doing this via the [`organize:`](https://snapcraft.io/docs/snapcraft-parts-metadata#organize) section work?

tomponline (approved):
Thanks!

simondeziel (approved):
Wasn't fast enough to provide feedback before the merge, but maybe check if `organize:` would work?

lxd-qemu-snap/snapcraft.yaml
mihalicyn: Can we create a symlink using the `organize` step? I haven't known about that.

simondeziel: No, it seems to only do renames.

mihalicyn: ah, then I think the `organize` thing is not very helpful in this case. We don't want to have `usr/lib` at the end. It's just a temporary hack to provide a cleanup script with what it expects to see. At the end we remove `usr/lib` symlink and all libraries live in `lib`.

PR #506, opened 2024-07-11
lxd: Cherry-picks (5.0-candidate)

tomponline:


MusicDin (commented):
Should also gorilla websocket be reverted to 1.5.1 (from 1.5.3)?

tomponline:
> Should also gorilla websocket be reverted to 1.5.1 (from 1.5.3)?

Yes in stable-5.0 but not here because its based on lxd-5.0.3 which used 1.5.1 https://github.com/canonical/lxd/blob/lxd-5.0.3/go.mod#L20

https://github.com/canonical/lxd/pull/13739

MusicDin (approved):

PR #510, opened 2024-07-12
Invoke virt-v2v-in-place on host

MusicDin:
This is required for virt-v2v-in-place to be invoked on the host.

Requires: https://github.com/canonical/lxd/pull/13748


tomponline (approved):

simondeziel:
It's a bit late now (I was in vacation) but could we have used the existing `run-host` wrapper mechanism? Are the `unset XDG_RUNTIME_DIR` and `PYTHONPATH` the reason why it was forked into another wrapper?

tomponline:
If we can use run-host that would be useful.

PR #516, opened 2024-07-15
snapcraft: make (dis)connect-plug-qemu-external privileged

mihalicyn:
Everything was working locally, likely because I was using --devmode. But for LXD installed from store we need to explicitly give necessary privileged to these hooks.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -179,6 +179,14 @@ hooks:
     plugs:
       - lxd-support
       - system-observe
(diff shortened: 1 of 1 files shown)


simondeziel (approved):

tomponline:
@mihalicyn for future reference please avoid using `--devmode` and instead use `--dangerous` as this still uses snapd's apparmor profile, it just doesn't require signatures.

tomponline (approved):

PR #538, opened 2024-08-21
Invoke virt-v2v-in-place on host (5.21-edge)

MusicDin:
From https://github.com/canonical/lxd-pkg-snap/pull/510

This is required for conversion option `virtio` which failed the tests for 5.21-edge in https://github.com/canonical/lxd-ci/pull/274.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1496,6 +1496,7 @@ parts:
         -not -path "${CRAFT_PRIME}/bin/remote-viewer" \
         -not -path "${CRAFT_PRIME}/bin/snap-query" \
         -not -path "${CRAFT_PRIME}/bin/sshfs" \
+        -not -path "${CRAFT_PRIME}/bin/virt-v2v-in-place" \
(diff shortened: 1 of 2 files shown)


tomponline:
@MusicDin are you planning to switch latest/edge over so that snapcraft/wrappers/virt-v2v-in-place uses run-host internally?

MusicDin:
As per our discussion, I've understood that we will keep the wrapper function to avoid adding exceptions in the `run-host`. To be specific, we need to unset the XDR_RUNTIME_DIR only in case of virt-v2v-in-place.

MusicDin:
Ahh sorry, I see. I'll update the wrapper to unset the variable and call the `run-host`.

simondeziel:
As found out by @MusicDin, trying to call `run-host` would require creating a symlink in the host's FS as `run-host` assumes `$0` to refer to actual command. As such, I think we have to live with 2 very similar wrapper scripts like this PR already implements. Thanks for investigating.

simondeziel (approved):

PR #544, opened 2024-09-05
Fix for cobra shell completions

kadinsayani:


simondeziel (dismissed):

snapcraft.yaml
```diff
@@ -1448,7 +1448,7 @@ parts:
 
       # Setup bash completion
       mkdir -p "${CRAFT_PART_INSTALL}/etc/bash_completion.d/"
-      "${CRAFT_PART_INSTALL}/bin/lxc" completion bash > "${CRAFT_PART_INSTALL}/etc/bash_completion.d/snap.lxd.lxc"
+      "${CRAFT_PART_INSTALL}/bin/lxc" completion bash |  sed '/\bcomplete\b/ s/lxc$/lxc lxd.lxc/' | sed 's/requestComp="${words\[0\]} __complete ${args\[\*\]}"/requestComp="\$SNAP\/lxc __complete ${args[*]}"/' > "${CRAFT_PART_INSTALL}/etc/bash_completion.d/snap.lxd.lxc"
```
tomponline: extra space before sed

simondeziel: A single `sed` also works using `; ` separator. An alternative would also be a single `sed` but with 2 edits like `sed -e 's/foo/bar/' -e 's/bla/boo/'`.
```suggestion
      "${CRAFT_PART_INSTALL}/bin/lxc" completion bash | sed '/\bcomplete\b/ s/lxc$/lxc lxd.lxc/; s/requestComp="${words\[0\]} __complete ${args\[\*\]}"/requestComp="\$SNAP\/lxc __complete ${args[*]}"/' > "${CRAFT_PART_INSTALL}/etc/bash_completion.d/snap.lxd.lxc"
```

I'd also appreciate a quick comment as to what that edit/mangling achieves or fixes.

snapcraft.yaml
```diff
@@ -1447,8 +1447,10 @@ parts:
       fi
```
simondeziel: Hmm, seeing `CRAFT_STAGE` here feels wrong as we are in the build step.

masnax (dismissed):
I did get it working on Friday, but I also learned a few things:

* The installing packages from `ppa:ubuntu-lxc/daily` adds a file at `/usr/share/bash-completion/completions/_lxc` which interferes with the snap's completion for `lxc`. 

* `ubuntu-minimal:24.04` seems to never do completions properly, even with `bash-completion` installed. Instead `complete -p lxd.lxc` always returns `_minimal`. 

* A non-root user must be logged in because the snap completion script depends on `/usr/share/bash-completion/bash_completion` which is only loaded by `/etc/profile.d/bash_completion.sh` if `XDG_CONFIG_HOME` is set.

However, on `ubuntu:24.04`, with `apt update ; apt upgrade ; apt install bash-completion`, the completion works if you do the following. I'm not sure if there's a clean way to incorporate this into the actual script generation?

```bash
      # Setup bash completion
      mkdir -p "${CRAFT_PART_INSTALL}/etc/bash_completion.d/"

      # Snapd requires the unaliased command `lxd.lxc` to be supplied as the first command for completion to be detected.
      # So that completions applies to the alias `lxc` too, we need to include both.
      set_cmds='s/^\s*complete.*__start_lxc /&lxd.lxc /'

      # When executed by Snapd, the COLUMNS shell value is unset, so use $(tput cols) instead.
      set_cols='s/COLUMNS/$(tput cols)/'

      # The script is run from the snap environment, which does not have /snap/bin in its PATH.
      # However, the script will try to run `eval lxd.lxc` so we need to explicitly append the path.
      set_path='s|out=$(eval "|&/snap/bin/|'

      "${CRAFT_PART_INSTALL}/bin/lxc" completion bash | sed -e "${set_cmds}" -e "${set_cols}" -e "${set_path}" > "${CRAFT_PART_INSTALL}/etc/bash_completion.d
/snap.lxd.lxc"
```

kadinsayani:
> However, on `ubuntu:24.04`, with `apt update ; apt upgrade ; apt install bash-completion`, the completion works if you do the following. I'm not sure if there's a clean way to incorporate this into the actual script generation?

Thanks for helping out with this and sharing your findings! This helps a lot. 

The cobra built-in completions command uses the following to generate the script: https://github.com/spf13/cobra/blob/78bfc837fe358c750faa7e7f0a8016b300044d58/bash_completionsV2.go#L31

We could implement our own version of `genBashComp()` and our own [custom completions command](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md), but would it be worth the effort? cc. @tomponline

Alternatively, we can just continue with the approach of making the changes to the script using `sed`.

kadinsayani:
The latest changes to the commit include @masnax's suggestions and the usage of `share/bash-completion/completions`. From the [bash-completion docs](https://github.com/scop/bash-completion):

> In bash-completion >= 2.12, we search the data directory of bash-completion under the installation prefix where the target command is installed. When one can assume that the version of the target bash-completion is 2.12 or higher, the completion script can actually be installed to $PREFIX/share/bash-completion/completions/ under the same installation prefix as the target program installed under $PREFIX/bin/ or $PREFIX/sbin/. For the detailed search order, see also "Q. What is the search order for the completion file of each target command?" below.

tomponline:
@kadinsayani is this ready for review?

kadinsayani:
> @kadinsayani is this ready for review?

Not yet - I haven't been able to get completions working consistently across my local test environments.

tomponline:
> > @kadinsayani is this ready for review?
> 
> Not yet - I haven't been able to get completions working consistently across my local test environments.

OK thanks - what is the outstanding issue?

masnax:
> OK thanks - what is the outstanding issue?

I've been testing this as well and the main issue is just that the snap completions are really finicky. 

Here are the issues I'm still seeing:
* If you install LXD before installing `bash-completion` with `apt-get install bash-completion`, the completions will not work until you re-install LXD. This exists both in `5.21/stable` and the changes in this PR. This seems like a separate issue to this PR so not a blocker. (This is the issue I was seeing before with `ubuntu-minimal` only resolving `_minimal` as the completer for `lxc`).

* When testing in VMs (or anything with AppArmor), AppArmor interferes with the newer script because it tries to run `eval /snap/bin/lxd.lxc` but does not have permission to do so in the snap without `aa-exec -p unconfined`. So that we won't have to add another `sed` case, it would be nice if we could add `lxd.lxc` to `/snap/lxd/current/bin` so it can be executed in the snap like `lxc` can.

kadinsayani:
@tomponline, should be good to go 👍 

I've tested the snap in both a container and VM running ubuntu:24.04, and completions are working as expected, without the need to reinstall `bash-completion`. The only caveat is that completions only work when running `lxd.lxc`, but I think this is expected when using a debug snap installed using `snap install --dangerous lxd.snap`. I did try running `snap alias lxd.lxc lxc`, but creating the alias after installing the snap leaves `lxc` with `_minimal` completions. cc. @masnax

snapcraft.yaml
kadinsayani: I believe `CRAFT_STAGE` is being used here because there is a check in `strip:override-prime` which primes the documentation only if `arch != riscv64`. I think this check can be moved to the `lxd` build step though, and then we can just prime the documentation there. I'm still learning about snap though, so I may be off here :)

tomponline (commented):
I like the comments on those, thanks!

@simondeziel are you happy with this?

@kadinsayani does the auto complete work for `lxc` commands when using `sudo snap alias lxd.lxc lxc`

snapcraft.yaml
simondeziel: Sorry for not closing the loop here, I dealt with it in https://github.com/canonical/lxd-pkg-snap/pull/545

simondeziel:
>     * When testing in VMs (or anything with AppArmor), AppArmor interferes with the newer script because it tries to run `eval /snap/bin/lxd.lxc` but does not have permission to do so in the snap without `aa-exec -p unconfined`. So that we won't have to add another `sed` case, it would be nice if we could add `lxd.lxc` to `/snap/lxd/current/bin` so it can be executed in the snap like `lxc` can.

Adding a symlink in our snap is pretty cheap so why not!

snapcraft.yaml
```diff
@@ -1469,7 +1476,7 @@ parts:
 
       - lib/*/libidn.so.*
 
-      - etc/bash_completion.d/snap.lxd.lxc
+      - etc/bash_completion.d/snap.lxd.lxc
```
simondeziel: Trailing whitespace.

kadinsayani:
> @kadinsayani does the auto complete work for `lxc` commands when using `sudo snap alias lxd.lxc lxc`

Yes, we're good to go 👍 

I've updated the commit to add the unaliased `lxd.lxc` binary to `/snap/lxd/current/bin`. Thanks @masnax and @simondeziel  for your help with debugging this one and appreciate the patience.

simondeziel:
> Thanks @masnax and @simondeziel for your help with debugging this one and appreciate the patience.

I appreciate your persistence! Thanks

simondeziel (dismissed):

tomponline (approved):
Thanks!

snapcraft.yaml
```diff
@@ -1447,8 +1447,15 @@ parts:
       fi
 
       # Setup bash completion
+      # Snap completer requires access to unaliased lxc binary
+      cp "${CRAFT_PART_INSTALL}/bin/lxc" "${CRAFT_PART_INSTALL}/bin/lxd.lxc"
```
tomponline: i dont think we should be doing this.

tomponline (changes requested):
im not sure about that copying of `lxc` to `lxd.lxc`, this shouldnt be needed as its what snapd does for us

snapcraft.yaml
simondeziel: If I understood Max's comment, this is needed for `snapd` to be able to escape the Apparmor profile.

I thought we'd go with a symlink rather than a full copy though.

kadinsayani:
> im not sure about that copying of `lxc` to `lxd.lxc`, this shouldnt be needed as its what snapd does for us

The snap completer always passes in the unaliased `lxd.lxc` to our generated completions script. The completions script then calls something like `lxd.lxc __complete ${args[*]}` from within the snap environment which doesn't have access to `/snap/bin`. That's why we're including `lxd.lxc` in `/snap/lxd/current/bin`. Another option would be to add something like `export PATH=$PATH:/snap/bin`, or perhaps there is a better way to include the path in the build step?

snapcraft.yaml
tomponline: @kadinsayani please can you scheduler a meeting with yourself @simondeziel and @masnax to discuss this, as I need to understand why the old one worked and why this one needs all these, dare I say it, "hacks".

kadinsayani:
I've updated the stream editing to replace `requestComp="${words[0]} __complete ${args[*]}"` with `requestComp="/snap/lxd/current/bin/lxc __complete ${args[*]}"`, which follows the same approach as the old script. I've tested on a container and VM using the following:

```
snapcraft

lxc launch ubuntu:24.04 v1

lxc file push lxd_git-...snap c1/root/lxd.snap

lxc exec v1 -- su -

snap install lxd --channel latest/edge

snap install --dangerous /root/lxd.snap

exit
lxc exec v1 -- su -

sudo snap alias lxd.lxc lxc

lxc [TAB]
complete -p lxc
```

Completions work as expected ✅

simondeziel:
@kadinsayani in your test steps, was the `complete -p lxc` step required?

kadinsayani:
> @kadinsayani in your test steps, was the `complete -p lxc` step required?

It's not required but it should output `complete -F _complete_from_snap lxc` after you've set up the alias and run `lxc [TAB] ^C`.

masnax:
> @kadinsayani in your test steps, was the `complete -p lxc` step required?

This is just used to verify the completions are being loaded from the snap. 

As per these docs: https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html
```
If the -p option is supplied, or if no options are supplied, existing completion specifications are printed in a way that allows them to be reused as input. 
```

simondeziel:
> > @kadinsayani in your test steps, was the `complete -p lxc` step required?
> 
> This is just used to verify the completions are being loaded from the snap.
> 
> As per these docs: https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html
> 
> ```
> If the -p option is supplied, or if no options are supplied, existing completion specifications are printed in a way that allows them to be reused as input. 
> ```

Thanks, TIL.

tomponline (approved):
Thanks!

PR #564, opened 2024-10-08
LXD stop hook wrapper

tomponline:
This requires support in `lxd-user` to add support for the `callhook` subcommand.

Related to https://github.com/canonical/lxd/issues/13331

Depends on https://github.com/canonical/lxd/pull/14253

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1602,6 +1602,7 @@ parts:
     organize:
       commands/snap-query: bin/
       hooks/: snap/hooks/
+      wrappers/lxd-stophook: bin/lxd
       wrappers/gpu-2404-custom-wrapper: bin/
       wrappers/editor: bin/
       wrappers/remote-viewer: bin/
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1401,7 +1401,12 @@ parts:
 
       # Build the binaries
       go build -trimpath -o "${CRAFT_PART_INSTALL}/bin/lxc" github.com/canonical/lxd/lxc
-      go build -trimpath -o "${CRAFT_PART_INSTALL}/bin/lxd" -tags=libsqlite3 github.com/canonical/lxd/lxd
+
+      # Build LXD server binary into ${CRAFT_PART_INSTALL}/sbin/lxd so that it does not conflict with the
+      # lxd-stophook wrapper script which is stored in ${CRAFT_PART_INSTALL}/bin/lxd.
+      # This way when a container stops it will call "/snap/lxd/current/bin/lxd callhook" which is handled by the
+      # lxd-stophook script, which in turn will execute "/snap/lxd/current/bin/lxd-user callhook" to notify LXD.
+      go build -trimpath -o "${CRAFT_PART_INSTALL}/sbin/lxd" -tags=libsqlite3 github.com/canonical/lxd/lxd
 
       # Build static binaries
       CGO_ENABLED=0 go build -trimpath -o "${CRAFT_PART_INSTALL}/bin/lxd-agent" -tags=agent,netgo github.com/canonical/lxd/lxd-agent
@@ -1440,7 +1445,7 @@ parts:
       usr/share/misc/: share/misc/
       var/lib/usbutils/usb.ids: share/misc/
       usr/sbin/: bin/
-      sbin/: bin/
+      sbin/sgdisk: bin/
     prime:
       - bin/dnsmasq
       - bin/getfattr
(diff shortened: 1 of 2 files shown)


snapcraft/commands/daemon.start
```diff
@@ -652,8 +652,8 @@ if [ -n "${PID}" ] && [ "$(readlink "/proc/self/ns/mnt")" != "$(readlink "/proc/
 fi
 
 # LXD
-## Check for existing LXDs
-for pid in $(pgrep --euid 0 --full "lxd(\.debug)? --logfile"); do
+## Check for existing LXDs (check for bare lxd and lxd-daemon processes not running via /bin/sh, as well as lxd.debug processes).
+for pid in $(pgrep --euid 0 --full "(^lxd-daemon|^lxd|lxd.debug) --logfile"); do
```
simondeziel: Optional change:

```suggestion
for pid in $(pgrep --euid 0 --full "^lxd(|-daemon|\.debug) --logfile"); do
```
Which reads more naturally to me? Starts with `lxd` followed by nothing or `-daemon` or `.debug` suffix followed by `--logfile`.

simondeziel (commented):
While I see no technical issue with the PR, I'd rather avoid us moving away from `lxd` as the exec name for our daemon.

Unfortunately, I don't have anything better to propose at the time :/

snapcraft/commands/daemon.start
tomponline: the `^` bit must only be for lxd-daemon and lxd but not lxd.debug because of the way its invoked it will be prefixed with the path.

simondeziel: Very true, thanks! Then the only missing bit is escaping the `.` of `.debug` (`\.debug`).

tomponline:
> While I see no technical issue with the PR, I'd rather avoid us moving away from `lxd` as the exec name for our daemon.

Me too. Ive got a (simple) idea, will try in the morning.

tomponline:
The solution was to put the lxd binary in `/sbin` and the stop hook wrapper script in `/bin`.

lxd-stophook-wrapper/lxd
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# Use exec so that this script process is replaced.
+# This avoids polluting the process tree with this wrapper script.
+if [ "$1" = "callhook" ]; then
+    exec /snap/lxd/current/bin/lxd-user "$@"
+else
+    exec lxd "$@"
+fi
```
simondeziel: Just a question of style but elsewhere, we put the second `exec` outside of any `if/else` clause as the one inside the `if` is terminal anyway. https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft/wrappers/run-host#L10-L14 is one such example.

simondeziel (dismissed):

lxd-stophook-wrapper/lxd
tomponline: yeah lets do that

snapcraft/wrappers/lxd-stophook
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# Use exec so that this script process is replaced.
+# This avoids polluting the process tree with this wrapper script.
+if [ "$1" = "callhook" ]; then
+    exec /snap/lxd/current/bin/lxd-user "$@"
+else
```
simondeziel: `s/else/fi/` :)

simondeziel (approved):

snapcraft/wrappers/lxd-stophook
tomponline: I fixed this but must not have pushed it after hitting the issue with the snapd unconfined rule changes

snapcraft.yaml
```diff
@@ -1602,6 +1612,7 @@ parts:
     organize:
       commands/snap-query: bin/
       hooks/: snap/hooks/
+      wrappers/lxd-stophook: bin/lxd
```
simondeziel, line 1615: I missed that change in my previous review. I much prefer it to the previous extra part so thanks!

simondeziel (approved):

mihalicyn (approved):

snapcraft/wrappers/lxd-stophook
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# Use exec so that this script process is replaced.
+# This avoids polluting the process tree with this wrapper script.
+if [ "$1" = "callhook" ]; then
+    exec /snap/lxd/current/bin/lxd-user "$@"
+fi
+
+exec lxd "$@"
```
simondeziel, line 8: @tomponline It just occurred to me that we may not want to fallback to invoking `lxd` (and relying on `$PATH`) for the non-callhook scenario.

If we ended up in that `lxd-stophook` script for a non-callhook, I think it'd be cleaner to `exit 1` with an error message explaining what's wrong.

tomponline: noted, thanks, will try this!

PR #584, opened 2024-10-24
Snap tweaks

simondeziel:
This slightly reduces the overhead of the `lxc` command. In this rough benchmark, `lxc help >/dev/null` is used as it is self-contained (makes no API call).

Baseline:

```
$ timeout 30 bash -c 'i=0; trap "echo \$i" EXIT; while i=$((i+1)); do lxc help >/dev/null; done'
598
```

PR changes applied:

```
$ sudo mount -o bind,ro snapcraft/commands/lxc /snap/lxd/current/commands/lxc
$ timeout 30 bash -c 'i=0; trap "echo \$i" EXIT; while i=$((i+1)); do lxc help >/dev/null; done'
620
```

This is a meager improvement (~3.5%) but it moves the needle in the right direction. From ~19.93 execs/second to ~20.67 execs/second.

For comparison's sake, here's what I get from a simple **non-snap** binary (~1192 execs/second):

```
$ timeout 30 bash -c 'i=0; trap "echo \$i" EXIT; while i=$((i+1)); do /usr/bin/echo foo >/dev/null; done'
35767
```

### DIFF
--- a/snapcraft/commands/daemon.reload
+++ b/snapcraft/commands/daemon.reload
@@ -11,4 +11,4 @@ fi
 
 echo reload > "${SNAP_COMMON}/state"
 read -r PID < "${SNAP_COMMON}/lxd.pid"
-kill "$PID"
+exec kill "$PID"
--- a/snapcraft/commands/lxc
+++ b/snapcraft/commands/lxc
@@ -11,13 +11,6 @@ if [ -z "${SNAP_REAL_HOME:-""}" ]; then
     fi
 fi
 
-# Migrate data if needed
-if [ ! -d "${SNAP_USER_COMMON}/config" ]; then
-    if [ -d "${SNAP_REAL_HOME}/.config/lxc" ]; then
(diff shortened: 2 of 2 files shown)


simondeziel:
@tomponline this was tested with `lxc config edit v1`, and `lxc console --type=vga v1` as well as regular `lxc` stuff.

tomponline:
What do you think is the main cause of the remainer of the slow down? The aa-exec call or the wrapper script itself?

simondeziel:
> What do you think is the main cause of the remainer of the slow down? The aa-exec call or the wrapper script itself?

I've replaced the wrapper script with just:
```
$ cat /snap/lxd/current/commands/lxc
#!/bin/sh
exec echo help

$ sudo mount -o bind,ro snapcraft/commands/lxc /snap/lxd/current/commands/lxc
$ lxc
help
```

and still get awful results:

```
$ timeout 30 bash -c 'i=0; trap "echo \$i" EXIT; while i=$((i+1)); do lxc help >/dev/null; done'
704
```

So most of the overhead seems to be from "entering the snap confinement".

tomponline (approved):

PR #593, opened 2024-11-06
snapcraft: edk2: fix subhook submodule url

mihalicyn:
Also, we don't need cmocka submodule url fixup anymore as: https://github.com/tianocore/edk2/commit/2ad22420a710dc07e3b644f91a5b55c09c39ecf3

Suggested-by: Thomas Parrott <thomas.parrott@canonical.com>

PR #594, opened 2024-11-06
snapcraft: qemu: do git clone manually (5.0-edge)

mihalicyn:
Do git clone manually to prevent recursive git clone failure while pulling edk2/subhook submodule.

See also https://github.com/canonical/lxd-pkg-snap/pull/593

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -751,7 +751,6 @@ parts:
       - spice-protocol
       - spice-server
     source: https://github.com/qemu/qemu
-    source-commit: 6bbce8b464206e6622216b62841cb3e953d56eb8 # v8.0.5
     source-depth: 1
     source-type: git
     plugin: autotools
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -751,7 +751,6 @@ parts:
       - spice-protocol
       - spice-server
     source: https://github.com/qemu/qemu
-    source-commit: 6bbce8b464206e6622216b62841cb3e953d56eb8 # v8.0.5
```
simondeziel, line 754: Could we keep that but add `source-submodules: []` to avoid pulling submodules automatically?
See https://snapcraft.io/docs/snapcraft-yaml-schema for details.

tomponline (approved):
Excellent thanks!

snapcraft.yaml
tomponline: Yeah worth a try indeed but ill merge this now to unblock us

snapcraft.yaml
```diff
@@ -817,7 +816,26 @@ parts:
...
+      set -ex
+
+      # We have to do this manual git clone because of edk2/subhook submodule URL has changed.
+      # See https://github.com/tianocore/edk2/commit/95d8a1c255cfb8e063d679930d08ca6426eb5701
+      # What is interesting is that we don't even need to do:
+      # git config --global url.https://github.com/tianocore/edk2-subhook.git.insteadOf https://github.com/Zeex/subhook.git
+      # because QEMU version which we checkout points to that version of edk2 which does not depends on subhook submodule.
+      # And as we do `git submodule update --init --recursive` *after* `git checkout ..` we have no problems.
+      # While if we do the same with built-in functionality in snapcraft it fails, because snapcraft does
+      # `git clone --recurse-submodules ...` and pulls master branch and fails on edk2/subhook before snapcraft does
```
tomponline, line 828: Ah interesting thanks, this seems like something we should log as an improvement request with the snapcraft team.

PR #595, opened 2024-11-06
Rework how lvm.conf is patched

simondeziel:
Another attempt at fixing https://github.com/canonical/lxd/issues/14341

This changes how `lvm.conf` is patched to be done at build time rather than at `daemon.start` time. This implies that now, executables are using `/snap/lxd/current` instead of `/snap/lxd/${rev}`. `/etc/lvm/profile` files are now included in the snap. It's unclear if they are used or not but they are so small (11KiB) that it doesn't hurt to ship them.

Lastly, the daemon visible `/etc/lvm` dir now contains 2 symlinks pointing to the read-only snap mount while still allowing the creation of the `/etc/lvm/archive` and `/etc/lvm/backup` directories backed by the `tmpfs` `/etc`:

```
# LD_LIBRARY_PATH=/snap/lxd/current/lib/:/snap/lxd/current/lib/x86_64-linux-gnu/:/snap/lxd/current/zfs-2.2/lib PATH=/snap/lxd/current/zfs-2.2/bin:/snap/lxd/current/bin:$PATH nsenter --mount=/run/snapd/ns/lxd.mnt -- ls -l /etc/lvm
total 0
drwx------ 2 root root 1860 Nov 14 19:54 archive
drwx------ 2 root root   40 Nov 14 19:54 backup
lrwxrwxrwx 1 root root   29 Nov 14 19:43 lvm.conf -> /snap/lxd/x1/etc/lvm/lvm.conf
lrwxrwxrwx 1 root root   28 Nov 14 19:43 profile -> /snap/lxd/x1/etc/lvm/profile
```

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -563,6 +563,7 @@ parts:
       - -bin/lvmetad
       - -bin/lvmpolld
       - etc/lvm/lvm.conf
+      - etc/lvm/profile/*
       - lib/*/device-mapper/*
       - lib/*/libaio.so*
       - lib/*/libdevmapper*
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -550,9 +550,14 @@ parts:
       usr/sbin/: bin/
     override-build: |-
       # Patch lvm.conf
+      # lvm.conf changes for lvm2 from 22.04/core22
       sed -i \
         -e "s%\(# \)\?obtain_device_list_from_udev = 1%obtain_device_list_from_udev = 0%" \
         -e "s%\(# \)\?cache_file_prefix = \"\"%cache_file_prefix = \"lxd\"%" \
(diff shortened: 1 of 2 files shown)


simondeziel:
@tomponline when you get a chance, I'm eager to see if that will improve the situation with `lvm` slowness with `latest/edge` :)

snapcraft.yaml
```diff
@@ -548,6 +548,20 @@ parts:
       sbin/: bin/
       usr/lib/: lib/
       usr/sbin/: bin/
+    override-build: |-
+      # Patch lvm.conf
```
tomponline, line 552: I'll merge this for now, but please can we expand this comment to explain what is happening here and why. Thanks

tomponline (approved):

PR #600, opened 2024-11-07
snapcraft: edk2 submodules fixes (5.0-candidate)

tomponline:
- Move downloading submodules into override-pull stage.
 - edk2: remove unnecessary fix subhook submodule url

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -312,14 +312,13 @@ parts:
       set -ex
       git clone https://github.com/tianocore/edk2 . -b edk2-stable202208
 
+      # Pull submodules after switching to source-commit
+      git submodule update --init --recursive
+
       git config user.email "noreply@lists.canonical.com"
       git config user.name "LXD snap builder"
     override-build: |-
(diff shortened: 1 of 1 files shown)


tomponline:
@mihalicyn @simondeziel its interesting that in 5.0-candidate, that is based on core20, we cant use the craftctl tool and so for the `qemu` section we are just doing `git clone`, but are not initializing the sub modules.

And this seems to be building fine.

Which begs the question whether we need to do:

https://github.com/canonical/lxd-pkg-snap/pull/599/commits/fd74e9bedbe8ef18d68da6d43913fd7bc935b95e#diff-e68b9afd67bb77dec8eda77796d3e0b784c7011203114996383eed6533be0cefR825

after disabling the automatic submodule resolution in snapcraft via `source-submodules: []`

Or whether we should be doing `git submodule update --init --recursive` in this PR for `5.0-candidate`.

mihalicyn:
> And this seems to be building fine.

probably it's because qemu repository also contains pre-built binaries for firmwares (in https://github.com/qemu/qemu/tree/master/pc-bios). So depending on the build configuration these submodules are not always required.

mihalicyn (approved):

PR #608, opened 2024-11-12
snapcraft: nvidia-container: Enable shallow clone (5.0-candidate)

tomponline:
To reduce the size of the git repo clone (~2GB) which is causing failures on riscv64.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -582,10 +582,11 @@ parts:
       - libseccomp
     source: https://github.com/NVIDIA/libnvidia-container
     source-commit: 4c2494f16573b585788a42e9c7bee76ecd48c73d # v1.16.1
-    # XXX: fails to build if using source-commit and source-depth
-    #source-depth: 1
+    source-depth: 1
     source-type: git
     plugin: make
+    build-environment:
+      - GIT_TAG: "1.16.1" # Enables source-depth: 1, should match git tag without "v" prefix.
     build-packages:
       - bmake
       - curl


simondeziel (approved):

MusicDin (approved):

snapcraft.yaml
```diff
@@ -582,10 +582,11 @@ parts:
       - libseccomp
     source: https://github.com/NVIDIA/libnvidia-container
     source-commit: 4c2494f16573b585788a42e9c7bee76ecd48c73d # v1.16.1
-    # XXX: fails to build if using source-commit and source-depth
-    #source-depth: 1
+    source-depth: 1
     source-type: git
     plugin: make
+    build-environment:
+      - GIT_TAG: "1.16.1" # Enables source-depth: 1, should match git tag without "v" prefix.
```
simondeziel, line 589: I wonder if it wouldn't be feasible to use `export GIT_TAG="$(git describe --tags --abbrev=0 | sed 's/^v//')"` in the override-build?

tomponline:
I'll also apply this to our other series.

snapcraft.yaml
tomponline: I initially tried doing `export GIT_TAG="1.16.1"` and it didnt work when using `snapcraftctl build` so looks like it resets the env

simondeziel: That's why I suggested having that done in the `override-build` section. Unless you've had that export put there already?

tomponline: yeah thats right it was there

tomponline: before the call to `snapcraftctl build`

PR #623, opened 2024-11-19
Dependency updates (latest-edge)

tomponline:


snapcraft.yaml
```diff
@@ -596,7 +589,7 @@ parts:
 
   nvidia-container:
     source: https://github.com/NVIDIA/libnvidia-container
-    source-commit: 63d366ee3b4183513c310ac557bf31b05b83328f # v1.17.1
+    source-commit: 63d366ee3b4183513c310ac557bf31b05b83328f # v1.17.2
```
simondeziel, line 601: I think we also need to update the GIT_TAG build env variable. This double accounting is quite annoying... but I know we don't have any better way ATM.

tomponline: good spot

tomponline: fixed

simondeziel (approved):

PR #624, opened 2024-11-19
Pre LXD 6.2 preparation (latest-candidate)

tomponline:
- Syncs changes from latest/edge that are not core24 specific.
- Updates LXD source-commit.
- Bumps some dependencies.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -51,6 +51,7 @@ description: |-
    - openvswitch.external: Use the system's OVS tools (ignores openvswitch.builtin) [default=false]
    - ovn.builtin: Use snap-specific OVN configuration [default=false]
    - ui.enable: Enable the web interface [default=true]
+   - zfs.external: Use the system's ZFS tools [default=false]
 
  For system-wide configuration of the CLI, place your configuration in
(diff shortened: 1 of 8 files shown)


snapcraft.yaml
```diff
@@ -247,12 +253,14 @@ parts:
...
+      - lib/*/libssh.so*
+      - lib/*/libtcmalloc.so*
+      - lib/*/libunwind.so*
 
   criu:
     source: https://github.com/checkpoint-restore/criu
-    source-depth: 1
-    source-tag: v3.19
+    source-commit: f8b14286b092853a4485813e1efd564109df9123 # v3.19
+    #source-depth: 1
```
simondeziel, line 263: The `source-depth` and git stuff for the `criu` part seems to be for `core24` only so you might want to drop those, according to the PR description.

tomponline: it builds ok on core22, and if https://github.com/canonical/lxd-pkg-snap/pull/623 builds ok on armhf ill bump this up to v4 and remove the patches anyway.

simondeziel (approved):

PR #626, opened 2024-11-19
Remove dependency on `uuidgen`

simondeziel:
`ovs-ctl` uses `uuidgen` only on the first startup when there is no `system-id.conf` file. After the first startup, the freshly created UUID is saved into that file for later use.

By generating the exact same kind and format of UUID ahead of time using `systemd-id128` (readily available from core base), we can remove that dependency on `uuidgen`.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -686,8 +686,6 @@ parts:
     autotools-configure-parameters:
       - --enable-ssl
       - --prefix=
-    stage-packages:
-      - uuid-runtime
     organize:
       sbin/: bin/
       usr/bin/: bin/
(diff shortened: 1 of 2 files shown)


tomponline (commented):
What project were you working on to come across this out of interest?

simondeziel:
> What project were you working on to come across this out of interest?

Saw that `uuid-runtime` package while working on https://github.com/canonical/lxd/pull/14497 and that jogged my memory about wanting to drop that snap dep :)

simondeziel:
This was tested successfully to generate and persist a UUID between snap restarts after doing `snap set lxd openvswitch.builtin=true`. The patched `daemon.start` was bind mount for the test and `/bin/false` was bind mount on top of `/snap/lxd/curent/bin/uuidgen` to make sure that binary wasn't used.

tomponline (approved):

PR #647, opened 2024-12-12
Sync latest-candidate to 5.21-edge (5.21-edge)

tomponline:
As part of the pre-LXD 5.21.3 release preparations I am syncing the snap changes from latest-candidate (6.2) into 5.21-edge, and then will sync to 5.21-candidate.


snapcraft.yaml
```diff
@@ -278,7 +287,10 @@ parts:
       [ "$(uname -m)" != "x86_64" ] && [ "$(uname -m)" != "aarch64" ] && [ "$(uname -m)" != "ppc64le" ] && [ "$(uname -m)" != "s390x" ] && [ "$(uname -m)" != "armv7l" ] && exit 0
       set -ex
 
-      make USERCFLAGS=-Wno-error=format-truncation criu
+      git config user.email "noreply@lists.canonical.com"
+      git config user.name "LXD snap builder"
```
simondeziel, line 291: Unneeded as we don't do any cherry-picking or such, but also harmless.

simondeziel (approved):

snapcraft.yaml
tomponline: thanks, yeah was trying to sync as much as possible.

PR #654, opened 2024-12-16
lxd: Cherry picks (latest-candidate)

tomponline:
Applies fix from https://github.com/canonical/lxd/pull/14659


MusicDin (approved):
Not sure why branch name error is appearing, but commit seems fine.

Maybe if you changed target branch later, the check was not run, because it is comparing against `latest-edge`.

PR #679, opened 2025-01-07
snapcraft/wrappers/gpu-2404-custom-wrapper: adjust XDG_DATA_DIRS env variable

mihalicyn:
We have to adjust XDG_DATA_DIRS variable to make nvidia-container-toolkit happy and find necessary configuration files. Actually, XDG_DATA_DIRS environment variable value we provided with is already correct (it's getting prepared for us in https://github.com/canonical/mesa-2404/blob/main/scripts/bin/gpu-2404-provider-wrapper.in and
https://git.launchpad.net/~canonical-kernel-snaps/canonical-kernel-snaps/+git/kernel-snaps-u24.04/tree/hooks/kernel-gpu-2404-provider-mangler?h=pc-components) but for some reason nvidia-container-toolkit library expects these path to be relative to driver root, which leads to a situation when driver root is getting appended twice to a path and files can not be found. One possible (and clean) solution would be to explicitly specify config files search path with nvcdi.WithConfigSearchPaths() but it doesn't work because of a bug https://github.com/NVIDIA/nvidia-container-toolkit/pull/847.

So the easiest way I found is to adjust XDG_DATA_DIRS variable in that weird way so when nvidia-container-toolkit prepend driverRoot to it a final path became valid and everything works.

This is to fix an issue with delivering some of the nvidia files that CDI does by default, which are the config files for things like Vulkan and EGL  The files are included in the kernels user components, so it just seems like LXD is excluding passing them through via CDI.

Please don't consider this an exhaustive list of files, necessarily, but examples are:

```
/usr/share/glvnd/egl_vendor.d/10_nvidia.json
/usr/share/glvnd/egl_vendor.d/50_mesa.json
/usr/share/vulkan/icd.d/nvidia_icd.json
/usr/share/vulkan/implicit_layer.d/nvidia_layers.json
/usr/share/egl/egl_external_platform.d/15_nvidia_gbm.json
/usr/share/egl/egl_external_platform.d/10_nvidia_wayland.json
```

### DIFF
--- a/snapcraft/wrappers/gpu-2404-custom-wrapper
+++ b/snapcraft/wrappers/gpu-2404-custom-wrapper
@@ -5,6 +5,12 @@
 if snapctl is-connected gpu-2404
 then
     echo "INFO: the gpu-2404 interface is connected. Running with gpu-2404 wrapper."
+    #
+    # We need this to help nvidia-container-toolkit with finding of
+    # some configuration files so they can be bindmounted in the container
+    # if needed.
+    #
+    export XDG_DATA_DIRS=${XDG_DATA_DIRS:+$XDG_DATA_DIRS:}/usr/share
     exec "${SNAP}/gpu-2404/bin/gpu-2404-provider-wrapper" "$@"
 else
     echo "INFO: the gpu-2404 interface isn't connected. Skipping gpu-2404 wrapper."


snapcraft/wrappers/gpu-2404-custom-wrapper
```diff
@@ -5,6 +5,7 @@
 if snapctl is-connected gpu-2404
 then
     echo "INFO: the gpu-2404 interface is connected. Running with gpu-2404 wrapper."
+    export XDG_DATA_DIRS=${XDG_DATA_DIRS:+$XDG_DATA_DIRS:}/usr/share
```
tomponline, line 13: Please can you add a comment here explaining why this is needed.

What is the "+" doing here btw?

mihalicyn: >What is the "+" doing here btw?

Hmm, I'm using the same way we have in many other places.
But strictly speaking, it is (from bash man pages):
```
       ${parameter:+word}
              Use Alternate Value.  If parameter is null or unset, nothing is substituted, otherwise
              the expansion of word is substituted.
```

mihalicyn: >Please can you add a comment here explaining why this is needed.

Have done.

tomponline: Thanks.

I'm struggling to see what this change is doing, or rather, why it helps.

You seem to be just adding XDG_DATA_DIRS to itself and then `/usr/share` too. Am I interpreting this correctly?

tomponline: Was `/usr/share` missing from the XDG_DATA_DIRS value set externally?

mihalicyn: >You seem to be just adding XDG_DATA_DIRS to itself and then /usr/share too. Am I interpreting this correctly?

Yeah, precisely.

>Was /usr/share missing from the XDG_DATA_DIRS value set externally?

Yep. It's not there and we add it.

All paths in `XDG_DATA_DIRS` are adjusted with a snap prefix like `/snap/lxd/<rev>/gpu-2404-2/<...>/usr/share` and this is correct. But nvidia-container-toolkit tries to append a  `driverRoot` (=`/snap/lxd/<rev>/gpu-2404-2`) value as a prefix (to each path from `XDG_DATA_DIRS`) which leads to a situation when these path became incorrect (something like `/snap/lxd/<rev>/gpu-2404-2/snap/lxd/<rev>/gpu-2404-2/usr/share`). By adding `/usr/share` I workaround this.

tomponline: Thanks for the info. And thanks for opening the upstream bug fix too.

tomponline (approved):

PR #688, opened 2025-01-16
Redirect `journalctl` to the host

gabrielmougard:
JIRA ticket: https://warthogs.atlassian.net/browse/LXD-1361
MM thread: https://chat.canonical.com/canonical/pl/9mmczm4kzbb18ebgraxyb69gga

LXD associated PR: https://github.com/canonical/lxd/pull/14805

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -233,6 +233,9 @@ fi
 # Redirect getent to the host
 ln -s "${SNAP}/wrappers/run-host" "/run/bin/getent"
 
+# Redirect journalctl to the host
+ln -s "${SNAP}/wrappers/run-host" "/run/bin/journalctl"
+
 # Redirect pro to the host
 ln -s "${SNAP}/wrappers/run-host" "/run/bin/pro"
 


tomponline (dismissed):
@simondeziel i cant think of a reason why we would need to use the journalctl inside the snap package can you? If not, then we should be ok to intercept calls to "journalctl" with this wrapper to execute the host's one in the host's namespace.

snapcraft.yaml
```diff
@@ -1597,6 +1597,7 @@ parts:
         -not -path "${CRAFT_PRIME}/bin/snap-query" \
         -not -path "${CRAFT_PRIME}/bin/sshfs" \
         -not -path "${CRAFT_PRIME}/bin/virt-v2v-in-place" \
+        -not -path "${CRAFT_PRIME}/bin/journalctl" \
```
simondeziel: If possible, this list should be kept in alphabetical order.

snapcraft.yaml
```diff
@@ -1653,3 +1654,4 @@ parts:
       wrappers/remote-viewer: bin/
       wrappers/sshfs: bin/
       wrappers/virt-v2v-in-place: bin/
+      wrappers/journalctl: bin/
```
simondeziel: @tomponline you suggested not using `run-host` for `journalctl` interception but didn't elaborate as to why you preferred a separated wrapper.

I think that having `daemon.start` do:

```
ln -s "${SNAP}/wrappers/run-host" "/run/bin/journalctl"
```

Would achieve the same and "shadow" `journalctl` only for LXD's snap. So same as what this separated wrapper would do but without duplicating it.

snapcraft.yaml

gabrielmougard:
@simondeziel updated

snapcraft.yaml
tomponline: Yes that was because at the time i didnt want to shadow journalctl inside the snap. But @gabrielmougard went ahead and did it anyway ;) and this gave me pause for thought as to whether that would cause any problems.

But i cant think of any off the top of my head, and so, yes, using a run-host symlink would be better.

simondeziel (dismissed):
@gabrielmougard please try to use the `run-host` with a symlink done by `daemon.start`. Similar to what's done for the `pro` client (see `/usr/bin/pro` for inspiration).
Thanks!

gabrielmougard:
@simondeziel updated

tomponline (approved):

PR #715, opened 2025-02-05
Workaround to ensure LXD is stopped before MicroCeph/MicroOVN (5.21-candidate)

simondeziel:


tomponline:
@simondeziel im going to hold off on this one for a bit until we've seen it in 6/stable, just in case there are any unforeseen issues.

tomponline (approved):

PR #722, opened 2025-02-07
Update nvidia-container/nvidia-container-toolkit parts to v1.17.4

simondeziel:


snapcraft.yaml
```diff
@@ -559,8 +559,6 @@ parts:
     source-depth: 1
     source-type: git
     plugin: make
-    build-environment:
```
tomponline, line 562: No it was needed to do shallow clone to avoid 2gb clone https://github.com/canonical/lxd-pkg-snap/commit/5c9aee86cb162393daa8a382d8f3ec2826c38425

tomponline: The build process tried to get git tag but wasnt available in shallow clone  so had to set it manually

simondeziel: That's right, sorry for not build testing this before submitting. My previous commit broke the build :/
The PR builds fine now.

tomponline (approved):

PR #723, opened 2025-02-07
Only install build packages on supported arches

simondeziel:


snapcraft.yaml
```diff
@@ -290,15 +290,18 @@ parts:
...
-      - libcap-dev
-      - libnet1-dev
-      - libnl-3-dev
-      - libprotobuf-c-dev
-      - libprotobuf-dev
-      - protobuf-c-compiler
-      - protobuf-compiler
-      - xmlto
+      - on riscv64:
+        - cmake      # dummy pkg (actually used by lxd part) to avoid installing build-packages on unsupported arches
```
tomponline: I dont understand what this is doing despite comment, why is it a dummy package?

simondeziel: Because that part is not going to be built on `riscv64`. Since I want to avoid repeating the long list for all the parts we build it for, I leverage a dummy pkg that I know is going to be required anyway (for the `lxd` part). This way, I can use the `else` clause and provide the list only once.

Is that clearer?

tomponline: Ah so this because you have to specify at least one package?

simondeziel: Yeah, as providing an empty list generates an error like in this made up example:

```
   logrotate:
     plugin: nil
+    build-packages:
+      - on amd64:
     stage-packages:
       - logrotate
     organize:
       usr/bin/: bin/
```

```
$ snapcraft
snapcraft internal error: 1 validation error for GrammarAwareProject
parts.logrotate.build-packages.0
  Value error, value must be a list of str: None [type=value_error, input_value={'on amd64': None}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error
```

An alternative hack that seems to fool the parser is to use something like this:

```
   logrotate:
     plugin: nil
+    build-packages:
+      - on amd64:
+        - ""
     stage-packages:
       - logrotate
     organize:
       usr/bin/: bin/
```

There, `snapcraft` doesn't complain.

tomponline: that latter option, with a comment explaining it, seems preferable rather than referencing a package used by a different part.

simondeziel: I had picked `cmake` because it was needed anyway on every arch but OK, I'll go with the empty variant.

tomponline (approved):

PR #754, opened 2025-03-07
snapcraft: explicitly add libtirpc-dev dev for zfs* parts

mihalicyn:
We have libtirpc-dev declared as a dependency for nvidia-container, but it is also required for zfs* parts.

It should fix a random failures like this one:
https://launchpadlibrarian.net/780529969/buildlog_snap_ubuntu_noble_s390x_lxd-latest-edge_BUILDING.txt.gz


tomponline (approved):
Thanks!

PR #755, opened 2025-03-10
Revert "snapcraft: bump snapd requirement to 2.67"

mihalicyn:
This reverts commit 98ab6204f970dba9aa7cd55384bba2c44adf0ff2.

... because of test failures
https://github.com/canonical/lxd-ci/actions/runs/13758467965

It turns out, that even on Ubuntu Noble LXD fails to install from latest/edge channel because snapd does not trigger autorefresh and just bail out LXD installation.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,7 +1,7 @@
 name: lxd
 base: core24
 assumes:
-  - snapd2.67
+  - snapd2.39
 version: git
 grade: devel
 summary: LXD - container and VM manager


mihalicyn:
Tom also pointed out, that we should look on the corresponding snapd package version in Debian.

On bookworm (stable) we have `2.57.6-1`:
https://packages.debian.org/bookworm/snapd

At the same time, users might run older Debian releases too. Without snapd autorefresh it's all extremely fragile.

tomponline (approved):
ta

simondeziel:
FYI, `snapd` 2.67.1 just landed in `-updates` on Ubuntu.

> Tom also pointed out, that we should look on the corresponding snapd package version in Debian.
> 
> On bookworm (stable) we have `2.57.6-1`: https://packages.debian.org/bookworm/snapd
> 
> At the same time, users might run older Debian releases too. Without snapd autorefresh it's all extremely fragile.

I don't think it's fair to wait for Debian because they seem to not include security fixes for `snapd`'s deb. Otherwise, they be shipping `2.63` already.

For the Debian crowd, I think we should just assume that `snapd` will have self-updated in time.

tomponline:
> FYI, `snapd` 2.67.1 just landed in `-updates` on Ubuntu.
> 
> > Tom also pointed out, that we should look on the corresponding snapd package version in Debian.
> > On bookworm (stable) we have `2.57.6-1`: https://packages.debian.org/bookworm/snapd
> > At the same time, users might run older Debian releases too. Without snapd autorefresh it's all extremely fragile.
> 
> I don't think it's fair to wait for Debian because they seem to not include security fixes for `snapd`'s deb. Otherwise, they be shipping `2.63` already.
> 
> For the Debian crowd, I think we should just assume that `snapd` will have self-updated in time.

My original point was for the snapd deb package in ubuntu not debian btw.

PR #759, opened 2025-03-12
snapcraft/qemu: Add QEMU patches for better error handling

gabrielmougard:
This should bring more visibility about what is failing when an error occurs in the `kvm_arch_get/put_registers` function calls.

These issues might arise when migrating a VM from CPU A to similar CPU B (even if we keep a common set of CPU flags between CPU A and B) Now, we should be able to see what the exact issue is.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -951,6 +951,12 @@ parts:
       # Apply patches from Ubuntu sources.
       QUILT_PATCHES=debian/patches quilt push -a
 
+      # Apply custom patches
+      # These patches enable more accurate error message returned in case of CPU registers mismatches.
+      # This is backported from QEMU v9.2.2.
+      patch -p1 < "${CRAFT_PROJECT_DIR}/patches/qemu-0001-kvm-Allow-kvm_arch_get-put_registers-to-accept-Error.patch"
+      patch -p1 < "${CRAFT_PROJECT_DIR}/patches/qemu-0002-target-i386-kvm-Report-which-action-failed-in-kvm_ar.patch"
+
       sed -i "s/^unset target_list$/target_list=\"${QEMUARCH}-softmmu\"/" configure
       sed -i 's#libseccomp_minver=".*#libseccomp_minver="0.0"#g' configure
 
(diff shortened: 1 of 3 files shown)


gabrielmougard:
@mihalicyn I ran `sudo snapcraft build qemu -v` successfully. But strangely, `sudo snapcraft build -v` blocks when building edk2 (I stopped around 2000s)

mihalicyn:
> But strangely, `sudo snapcraft build -v` blocks when building edk2 (I stopped around 2000s)

You can try to use a remote build https://github.com/canonical/lxd-pkg-snap?tab=readme-ov-file#build-the-lxd-snap-on-launchpad
```
snapcraft remote-build --launchpad-accept-public-upload --platform amd64
```

Then you can locally test it by doing something like:
```
snap remove --purge lxd
snap install lxd_5.15_amd64.snap --dangerous
snap alias lxd.lxc lxc
lxd init --auto
...
```

It's better to test this changes on practice to see if it give more verbose and helpful error messages on your experimental environment.

snapcraft.yaml
```diff
@@ -929,6 +929,10 @@ parts:
       # Apply patches from Ubuntu sources.
       QUILT_PATCHES=debian/patches quilt push -a
 
+      # Apply custom patches
```
simondeziel, line 954: We'll likely drop those patches when moving to `core26` but for consistency, I'd like to see a comment saying those patches are backported from QEMU version 9.X.Y. Just in case we move away from QEMU built from Noble sources.

tomponline (commented):
Did you say you had shrunk this PR @gabrielmougard still looks pretty large to me!

snapcraft.yaml
gabrielmougard: fixed

gabrielmougard:
@tomponline Now it should be good.

simondeziel (approved):

mihalicyn (approved):
Great work, @gabrielmougard !

LGTM

tomponline (approved):

mihalicyn:
https://launchpadlibrarian.net/784982718/buildlog_snap_ubuntu_noble_arm64_lxd-latest-edge_BUILDING.txt.gz

gabrielmougard:
@mihalicyn thanks for pointing that out... I'll try to see why this fails (it was not failing when I built the snap last time though...)

mihalicyn:
It's on arm64. You probably need to do `snapcraft remote-build --launchpad-accept-public-upload --platform arm64` to reproduce.

PR #760, opened 2025-03-12
libnvidia-container: prime `libnvidia-container-go.so` that is `dlopen`'ed

simondeziel:
As found by Tom while doing tests with NVIDIA cards, the `-go.so` file is needed as `libnvidia-container.so` will [`dlopen()` it](https://github.com/NVIDIA/libnvidia-container/blob/95d3e86522976061e856724867ebcaf75c4e9b60/src/nvcgo.c#L113).

That `dlopen()` is also why the snapcraft linter and my `ldd` tests didn't catch this runtime dependency.

Fixes bcb80f7d4dfb765dc14efeb774b45eeb707a056b

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -643,6 +643,7 @@ parts:
     prime:
       - bin/nvidia-container-cli*
       - lib/libnvidia-container.so*
+      - lib/libnvidia-container-go.so*  # dlopen()'ed by libnvidia-container.so
 
   nvidia-container-toolkit:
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline I went for a new commit rather than reverting the faulty commit as it allows being a bit more explicit about why it's needed.

Once merged, I'll cherry-pick into `latest-candidate`.

tomponline (approved):

tomponline:
Ta!

PR #769, opened 2025-03-20
lxd-qemu-snap: bump QEMU/virglrenderer versions

mihalicyn:
Bump QEMU to v9.2.2
virglrenderer to v1.1.0

Manually tested.

### DIFF
--- a/lxd-qemu-snap/snapcraft.yaml
+++ b/lxd-qemu-snap/snapcraft.yaml
@@ -56,7 +56,7 @@ parts:
 
   virgl:
     source: https://gitlab.freedesktop.org/virgl/virglrenderer.git
-    source-tag: virglrenderer-1.0.1
+    source-tag: virglrenderer-1.1.0
     source-depth: 1
     plugin: meson
     meson-parameters:
@@ -70,7 +70,7 @@ parts:
     after:
       - virgl
     source: https://gitlab.com/qemu-project/qemu
-    source-commit: 6a54d5cf55b446ec50d5c5a0b0568a7b28e2913e # v9.0.3
+    source-commit: ea35a5082a5fe81ce8fd184b0e163cd7b08b7ff7 # v9.2.2
     source-depth: 1
     source-submodules: []
     source-type: git


lxd-qemu-snap/snapcraft.yaml
```diff
@@ -56,7 +56,7 @@ parts:
 
   virgl:
     source: https://gitlab.freedesktop.org/virgl/virglrenderer.git
-    source-tag: virglrenderer-1.0.1
+    source-tag: virglrenderer-1.1.0
```
simondeziel, line 59: I don't remember if it was possible to use `source-commit` for this part? If so, could you use `1aeaf5e10a9c89096e96d09599aa419d5c50712f` which belongs to that tag?

simondeziel (approved):

lxd-qemu-snap/snapcraft.yaml
mihalicyn: To be honest I don't know why we are using `source-commit` thing for QEMU in there. We don't distribute this snap and never upload it to the store. It's more convenient to have explicit tag/branch references in this case, IMHO. This `snapcraft.yml` file is just a reference example and should not be considered as something with any security guarantees anyways.

mihalicyn: For some reason, Tom changed it to use `source-commit` thing in https://github.com/canonical/lxd-pkg-snap/commit/d1ec45f6ce657bedfe52ce450c621853b4abd9f4 commit.

simondeziel: I do agree that `source-tag` is more convenient but tags are mutable meaning we don't have guaranties that we are getting the same bits from one build to the other.

This snap being an unpublished one indeed changes how we need to be strict about it. I'm fine either way which is why I approved it as-is ;)

mihalicyn: I'm not against going with `source-commit` thing. Just decided to mention a convenience/applicability of `source-tag` in this specific case ;-)

PR #770, opened 2025-03-20
daemon.start: drop broken code around /dev/pts handling

simondeziel:
This code was introduced by 6edb1c4cbddf671e3211bf18c5fcf16a16afd4c0 However, `grep "devpts /dev/pts "`'s output can never match `"2"`.


simondeziel:
I can only guess the initial intention was to use `grep -c` and then take action when 2 hits were found. @mihalicyn maybe you'd prefer for the check to be fixed rather than be dropped?

tomponline:
Is this causing any issues btw?

simondeziel:
> Is this causing any issues btw?

I'd say no as it remained unfixed/inactive since its introduction in August 2020.

mihalicyn:
yep, looks like we can safely remove this.

mihalicyn (approved):

PR #786, opened 2025-04-01
patches/qemu: Fix QEMU patch for arm64 build

gabrielmougard:
This should fix the snap build for ARM64 with the QEMU patches which was failing due to a missing C function definition in the `arm/kvm64.c` backend.


gabrielmougard:
I have a working build using: `snapcraft remote-build --launchpad-accept-public-upload --build-for arm64`

tomponline (approved):

PR #805, opened 2025-05-14
Follow-up fixes for smaller `liblxc.so`

simondeziel:


snapcraft.yaml
```diff
@@ -1347,6 +1347,7 @@ parts:
       - ninja-build
     plugin: meson
     meson-parameters:
+      - "--buildtype=release"
```
mihalicyn, line 1350: Does this mean that we get no debug info in the resulting binary?

snapcraft.yaml
```diff
@@ -1361,6 +1361,7 @@ parts:
       - -Drootfs-mount-path=/var/snap/lxd/common/lxc/
       - -Dseccomp=true
       - -Dselinux=false
+      - -Dspecfile=false
```
tomponline, line 1365: please can you include why statements or impact statements in ur commit messages - i dont know what this does/doesnt do. Ta

snapcraft.yaml
simondeziel: We were already stripping the debug symbols in the `strip` (fake) part. It also influences the `-O` level changing from `-O0` to `-O3`, see https://mesonbuild.com/Builtin-options.html#details-for-buildtype which I also linked in the commit message now.

snapcraft.yaml
simondeziel: Sure, done.

tomponline (approved):

PR #806, opened 2025-05-17
Improve `lxd.buginfo` output

simondeziel:
The goal is to eventually advise bug reporters to provide the `sudo lxd.buginfo` output when filling out bugs for LXD.


tomponline (approved):
Ta

PR #807, opened 2025-05-19
External dependency updates (latest-edge)

tomponline:


markylaing (approved):
LGTM :+1: 

General comment - do we have a way of performing this automatically as we do for go dependencies etc.?

tomponline:
> LGTM 👍
> 
> General comment - do we have a way of performing this automatically as we do for go dependencies etc.?

Sadly not. Also its not a case of bumping to latest in the repo, in the cases when we use Ubuntu sources we only want to take from the Noble LTS updates.

PR #820, opened 2025-05-22
lxd-ui: update 0.17 tag (latest-candidate)

edlerd:
## Done

- fix(instance) ensure only one instance is in loading state, when
multiple are started/stopped from different projects in the all projects
view with the same name
- fix(instances) invalidate correct query key on instance migration
- fix(instances) ensure to clear panel params when instance name and
project are invalid
- fix(instances) avoid duplicate instances in other projects to match
bulk action permission checks

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1647,7 +1647,7 @@ parts:
     source: https://github.com/canonical/lxd-ui
     source-depth: 1
     source-type: git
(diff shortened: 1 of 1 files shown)


tomponline (approved):

tomponline:
@edlerd can the ui test suite be expanded to cover these scenarios (and the last PRs)?

PR #821, opened 2025-05-22
snapcraft/wrappers/nvidia-container-cli: remove unused core24 handling

mihalicyn:
This was introduced when we were working on Ubuntu Core + nvidia.runtime=true support, but then we decided to go with newer NVIDIA CDI thing and this change was never be useful. Let's just drop it.

It is not a bugfix. No functional changes intended.


simondeziel (approved):

tomponline (approved):

tomponline:
Ta!

PR #836, opened 2025-06-10
lxd-ui: enable build on riscv64 (latest-edge)

edlerd:


tomponline (approved):
ta

tomponline (changes requested):
We dont need the `override-prime` and `override-build` lines as the default is to run `craftctl default`

edlerd:
As discussed, [nodejs](https://snapcraft.io/node) is not available for riscv64, but installed as part of the ui build. So this will fail anyhow.

I wonder if this should directly be closed. Wdyt @tomponline ?

edlerd:
Requested riscv64 to be added as a nodejs architecture in https://github.com/nodejs/snap/issues/77

simondeziel:
> Requested riscv64 to be added as a nodejs architecture in [nodejs/snap#77](https://github.com/nodejs/snap/issues/77)

Thanks for that! Once done and available, I think we should replace all the overrides and just use a `build-snap` directive same as we deploy the `go` snap.

tomponline:
Closing for now.

PR #839, opened 2025-06-12
Use empty `build-packages` and `stage-packages` list instead of placeholder package

simondeziel:
As noted by @tomponline the kludge of providing an unrelated package to the `armhf` list was quite confusing and turned out to not be needed. I probably had tried an empty list but with this bogus syntax:

```
    stage-packages:
      - to armhf:
      - else:
        ...
```

and (wrongly) concluded that empty lists were not possible while in fact it just needs to be properly formatted:

```
    stage-packages:
      - to armhf: []
      - else:
        ...
```


simondeziel:
This built successfully on `armhf` and `amd64`.

tomponline (approved):
Yay! Thanks for picking this up.

PR #842, opened 2025-06-23
enable non-shallow clone for lxd

minaelee:
LXD documentation requires `sphinx-last-updated-by-git`, which requires a non-shallow clone to enable both sitemap `<lastmod>` and the `Last updated on <date>` notice on each docs page.


tomponline (approved):
Ta!

PR #843, opened 2025-06-23
criu: provide empty `stage-packages` list for `armhf`

simondeziel:
Even if that part is not shipped in `riscv64` builds, those packages are available.


tomponline (approved):
Thanks!

simondeziel:
@tomponline tested fine during remote-build tests (ignoring the failure on `amd64` for non-shallow git).

PR #845, opened 2025-06-24
Try to be more resilient on broken mount namespace

simondeziel:


simondeziel:
@tomponline this is NOT tested, yet. Does that capture your intention though?

Also, there are a few more occurrences where `snapctl get` is used but they are in `connect`/`disconnect` and GPU wrapper scripts where failure is already gracefully handled.

snapcraft/commands/snap-query
tomponline, line 1: Lets add to the commit message that we are doing this in case the symlink configured to this path fails during daemon.start

snapcraft/hooks/configure
```diff
@@ -33,9 +33,12 @@ get_bool() {
     return
 }
 
-# Don't fail if the mount namespace isn't properly setup yet
+# Try to fix an improperly setup mount namespace
 if [ ! -e /run/snapd-snap.socket ]; then
-    exit 0
```
tomponline, line 38: I wonder if we should just fail entirely, im not sure when the `configure` hook is called and why the snap socket shouldnt always be available.

simondeziel: So you'd like the old behaviour of exiting on missing socket but with a `rc=1`?

tomponline: I'd like to try it, and see if it works, as otherwise it feels like it could be silently masking issues.

tomponline:
> failure is already gracefully handled

I'd like to understand more about what that means, as if it results in the snaps not being connected, that'll likely cause issues down the line and may well have been clearer to fail when unable to retrieve the config.

simondeziel:
For the GPU wrapper, the script is invoked before `daemon.start` even started:

```
Jun 22 15:14:40 sdeziel-lemur systemd[1]: Started Service for snap application lxd.daemon.
Jun 22 15:14:40 sdeziel-lemur lxd.daemon[3012]: INFO: the gpu-2404 interface is connected. Running with gpu-2404 wrapper.
Jun 22 15:14:40 sdeziel-lemur lxd.daemon[3012]: => Preparing the system (34300)     # first thing done by `daemon.start`
```

So we've got a chicken-n-egg problem it seems.

tomponline:
Another reason to action https://github.com/canonical/lxd/issues/15835 then :)

tomponline:
> Another reason to action [canonical/lxd#15835](https://github.com/canonical/lxd/issues/15835) then :)

Although im surprised how it works then - must rely on a previous daemon start to setup the symlink

simondeziel:
Hmm, the GPU wrapper suffers for more problems as it does not escape from Apparmor:

```
Jun 20 08:43:51 sdeziel-lemur systemd[1]: Listening on Socket unix for snap application lxd.daemon.
Jun 20 08:43:52 sdeziel-lemur systemd[1]: Started Service for snap application lxd.daemon.
Jun 20 08:43:52 sdeziel-lemur lxd.daemon[1210390]: INFO: the gpu-2404 interface is connected. Running with gpu-2404 wrapper.
Jun 20 08:43:52 sdeziel-lemur audit[1210390]: AVC apparmor="DENIED" operation="open" class="file" profile="snap.lxd.daemon" name="/var/lib/snapd/hostfs/etc/nsswitch.conf" pid=1210390 comm="gpu-2404-provid" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Jun 20 08:43:52 sdeziel-lemur audit[1210390]: AVC apparmor="DENIED" operation="open" class="file" profile="snap.lxd.daemon" name="/var/lib/snapd/hostfs/etc/passwd" pid=1210390 comm="gpu-2404-provid" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Jun 20 08:43:52 sdeziel-lemur lxd.daemon[1210390]: => Preparing the system (34300)
```

It also has more issues during system shutdown/halt of the host:

```
Jun 20 14:18:51 sdeziel-lemur systemd[1]: snapd.service: Deactivated successfully.
Jun 20 14:18:51 sdeziel-lemur systemd[1]: Stopped Snap Daemon.
Jun 20 14:18:51 sdeziel-lemur systemd[1]: snapd.service: Consumed 1min 27.259s CPU time.
Jun 20 14:18:51 sdeziel-lemur systemd[1]: Stopped target System Time Set.
Jun 20 14:18:51 sdeziel-lemur lxd.daemon[1297711]: error: cannot communicate with server: Post "http://localhost/v2/snapctl": read unix @->/run/snapd-snap.socket: read: connection reset by peer
Jun 20 14:18:51 sdeziel-lemur lxd.daemon[1297637]: INFO: the gpu-2404 interface isn't connected. Skipping gpu-2404 wrapper.
Jun 20 14:18:51 sdeziel-lemur lxd.daemon[1297637]: => Stop reason is: host shutdown
Jun 20 14:18:51 sdeziel-lemur lxd.daemon[1297637]: => Stopping LXD (with instance shutdown)
```

tomponline:
@mihalicyn @simondeziel any ideas how the daily GPU tests pass if the wrapper doesnt work?

tomponline:
@simondeziel we may not actually need to run it during shutdown, it might just have been done like that due to the use of the chain command.

simondeziel:
> @mihalicyn @simondeziel any ideas how the daily GPU tests pass if the wrapper doesnt work?

Looking at fresh boot logs, it seems to just work on boot so maybe `snapd` is responsible for creating the symlink itself? Maybe LXD grew this capability just to paper over some unreliability?

tomponline:
Lets remove it and find out :)

snapcraft/hooks/configure
mihalicyn: > if we should just fail entirely, im not sure when the `configure` hook is called

I think that's a main question here. We probably need to consult with snapd code before changing this. I believe there was a reason for this silent masking, for instance, it can be that the `configure` hook is being called a few times during refresh on a different stages (where mount namespace is not yet ready). By returning an error we can break that.

mihalicyn:
> any ideas how the daily GPU tests pass if the wrapper doesnt work?

no, without wrapper tests must fail as they rely on the environment variables it sets up.

simondeziel:
This needs more investigation and probably some team discussion too so I'll close this half baked PR.

PR #881, opened 2025-08-07
snapcraft: cherry-pick `gomod: Update dependencies` (latest-candidate)

simondeziel:

PR #882, opened 2025-08-08
lxd: Update go dependencies with security fixes (latest-candidate)

tomponline:
Fixes build failure from https://github.com/canonical/lxd-pkg-snap/pull/881


skozina (approved):

simondeziel:
> Fixes build failure from #881

Thanks for the follow-up, I was offline so only just saw the fallout.

tomponline:
> > Fixes build failure from #881
> 
> Thanks for the follow-up, I was offline so only just saw the fallout.

No worries! :)

PR #901, opened 2025-08-28
lxd: cherry-pick: pure: Ensure image size is applied (5.21-candidate)

MusicDin:
Cherry-pick:
- https://github.com/canonical/lxd/pull/16339


tomponline (changes requested):
https://github.com/canonical/lxd/commit/2de60561bcd36c92110bf6c9b57594829c2e72c5 is in the main branch. As I mentioned in team standup yesterday we should only cherry pick commits from the associated lxd series branch (so in this case stable-5.21) to ensure we never cherry pick something and then it gets removed due to not being back ported in the next point release.

tomponline (approved):
Thanks!

PR #921, opened 2025-09-12
commands/daemon: redirect multipath tool to the host

mihalicyn:
Required for proper HPE Alletra iSCSI mode support.

I've already tested that this helps on my development environment. So it is good to be merged.


tomponline (approved):
ta

PR #934, opened 2025-10-02
liblxc: hardcode `libdir`

simondeziel:


tomponline (approved):

simondeziel:
Oh, I was just thinking twice and remembered that we do rely on the path to prime the `.so` into place. I didn't test with a local build due to being in meetings back to back... LP will soon let us know :/

PR #936, opened 2025-10-03
liblxc: fix the path in libdir

simondeziel:


simondeziel:
It built fine on `amd64`.

tomponline (approved):
ta

PR #938, opened 2025-10-07
Use `ln -snf` to create symlinks in OVN disconnect hooks

simondeziel:
This mimics Ceph hooks and also the OVN connect hooks.


tomponline (approved):
Ta

PR #944, opened 2025-10-14
daemon.start: make `/run/sysctl.d/zz-lxd.conf` world readable

simondeziel:
`mktemp` creates files with `0600` permissions which works fine with tools using it but since its content is not secret, it should be world readable.

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -483,6 +483,7 @@ manage_apparmor_restrictions() {
 
         # Save the overrides in a snippet file if content differs or any file is missing
         if ! cmp --quiet "${tmpf}" "${SYSCTL_FILE}"; then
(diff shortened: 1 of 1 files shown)


simondeziel:
Tested locally:

```
$ sudo mount -o ro,bind snapcraft/commands/daemon.start /snap/lxd/current/commands/daemon.start 
$ ls -l /run/sysctl.d/
total 0
$ lxc ls
...
$ ls -l /run/sysctl.d/
total 4
-rw-r--r-- 1 root root 104 Oct 14 12:29 zz-lxd.conf
$ cat /run/sysctl.d/zz-lxd.conf 
-kernel.apparmor_restrict_unprivileged_userns = 0
-kernel.apparmor_restrict_unprivileged_unconfined = 0
```

tomponline (approved):
Ta

PR #974, opened 2025-11-20
daemon.start: avoid using `cut` to extract ZFS MAJOR.MINOR version

simondeziel:


simondeziel:
Tested by locally bind mounting `daemon.start` into place.

snapcraft/commands/daemon.start
```diff
@@ -408,7 +408,8 @@ if [ "${zfs_external:-"false"}" = "false" ]; then
         VERSION=$(nsenter -t 1 -m modinfo -F version zfs 2>/dev/null || true)
     fi
 
-    ZFS_VER="$(echo "${VERSION}" | cut -c 1-3)"
+    # Extract MAJOR.MINOR from a string like 2.3.1-1ubuntu2
+    ZFS_VER="${VERSION%.*}"
```
simondeziel, line 412: The `cut` was working but flawed if the MAJOR or the MINOR was longer than 1 digit long:

```
$ echo 11.22.33 | cut -c 1-3
11.
```

While the variable substitution works better:

```
$ VERSION=11.22.33
$ echo "${VERSION%.*}"
11.22
```

tomponline (approved):

PR #975, opened 2025-11-21
daemon.start: don't manage Apparmor related sysctl if not active

### LINKED ISSUE #17012: lxd causes SELinux denial related to zz-lxd.conf file
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Fedora 

### Distribution version

42

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
lxd (candidate) 6.5-f33fdcf from Canonical** installed

Found in snapd CI.
```

### Output of "lxc info" or system info if it fails

```shell
-
```

### Issue description

The lxd daemon startup generates `zz-lxd.conf` as means of improving some interaction with AppArmor. However, the file is generated unconditionally, even if AppArmor is not enabled in the host's kernel.

In detail, on a SELinux system such as Fedora 42, with the latest policy, `/run` is labeled as `var_run_t`, in absence of type transition rules, files and directories created there will inherit the label. Since the reference policy does not carry any rules that would cover transition from `var_run_t` -> `system_conf_t` for files or directories when created by `unconfined_t` or `unconfined_service_t` we end up with default label. Startup script proceeds to trigger restart of `systemd-sysctl`, which runs under a policy that does not allow reading of `var_run_t` objects, thus triggering a denial:

```
 time->Fri Nov 21 06:24:09 2025
type=AVC msg=audit(1763706249.803:1865): avc:  denied  { read } for  pid=39914 comm="systemd-sysctl" name="zz-lxd.conf" dev="tmpfs" ino=6856 scontext=system_u:system_r:systemd_sysctl_t:s0-s0:c0.c1023 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=0
```

see snapd CI run for details: https://github.com/canonical/snapd/actions/runs/19558831113/job/56015065905#step:21:304

The change was introduced in https://github.com/canonical/lxd-pkg-snap/commit/79b82e48eebf80b6ee0aa6d68d19b2cc25a68b5f. The original code probed for presence of /proc/sys/kernel/apparmor_restrict_unprivileged_userns, which would not show up on a kernel without AppArmor. I'd be great if we could tweak the current behavior to match what fix for https://github.com/canonical/lxd/issues/16990 did and perhaps check for presence of /sys/kernel/security/apparmor before doing changes on the host. Other option is to run `restorecon -v -R /run/sysctl.d` before restarting systemd, but applying these sysctl changes makes no sense on those systems to begin with.

### Steps to reproduce

1. run Fedora 42
2. install lxd latest/candidate

### Information to attach

- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

simondeziel:
Closes https://github.com/canonical/lxd/issues/17012

Fixes regression introduced by 79b82e48eebf80b6ee0aa6d68d19b2cc25a68b5f

### DIFF
--- a/snapcraft/commands/daemon.start
+++ b/snapcraft/commands/daemon.start
@@ -458,6 +458,9 @@ manage_apparmor_restrictions() {
     # sysctl override snippet
     SYSCTL_FILE="/var/lib/snapd/hostfs/run/sysctl.d/zz-lxd.conf"
 
+    # Nothing to do if Apparmor is NOT active
+    [ -d /sys/kernel/security/apparmor ] || return
+
     # Check if sysctl settings need to be altered and always write the override snippet
     if [ "${apparmor_unprivileged_restrictions_disable:-"true"}" = "true" ]; then
         KEY_USERNS="kernel.apparmor_restrict_unprivileged_userns"


snapcraft/commands/daemon.start
```diff
@@ -458,6 +458,12 @@ manage_apparmor_restrictions() {
     # sysctl override snippet
     SYSCTL_FILE="/var/lib/snapd/hostfs/run/sysctl.d/zz-lxd.conf"
 
+    # If Apparmor is NOT active, ensure the sysctl snippet isn't left behind
+    if ! [ -d /sys/kernel/security/apparmor ]; then
+        rm -f "${SYSCTL_FILE}"
```
bboozzoo: AFAIU this code wasn't made available in stable yet, so it's unlikely that anyone would have the file (besides it goes away during reboot)
```suggestion
```

simondeziel: Thanks, done.

tomponline (approved):
ta

PR #977, opened 2025-11-21
Fix ZFS version extraction

simondeziel:


simondeziel:
@tomponline the previous version worked on my host with ZFS version `2.3.1-1ubuntu2` but failed with versions like `2.2.2-0ubuntu9.4`. The difference was the `9.4` bit which tripped the string substitution. So insufficient testing on my side :/

This change reintroduces `cut` but with proper splitting on `.` instead of counting chars. This should cope with eventual multi-digit MAJOR and/or MINOR numbers.

tomponline (approved):
ta

PR #989, opened 2025-11-26
snapcraft/commands/daemon.start: fix startup on non AppArmor systems

bboozzoo:
Fix startup on systems that have no AppArmor support.


tomponline (approved):

simondeziel (approved):
Sigh, thanks!

PR #991, opened 2025-11-26
Always `return` something in `daemon.start`

simondeziel:


tomponline (dismissed):
Do these need to go into LXD 6.6?

simondeziel:
> Do these need to go into LXD 6.6?

The first one only. But I've been thinking of simplifying the `get_bool` helper that is duplicated everywhere so maybe I should split this into 2 PRs and just deal with `daemon.start` here which would best land in 6.6 and the rest would target just `latest/edge`.

simondeziel:
> I should split this into 2 PRs and just deal with `daemon.start` here which would best land in 6.6 and the rest would target just `latest/edge`.

I've done just that.

tomponline (approved):

PR #1021, opened 2026-01-08
Add AMD CTK wrapper for snap environment

dmitry-lyfar:
Complements https://github.com/canonical/lxd/pull/17343.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -201,6 +201,16 @@ hooks:
 
 parts:
   # Dependencies
+  amd-ctk:
+    source: https://github.com/ROCm/container-toolkit
+    source-depth: 1
+    source-type: git
+    plugin: go
+    build-snaps:
+      - go
+    prime:
+      - bin/amd-ctk
+
   btrfs:
     plugin: nil
     stage-packages:


snapcraft/wrappers/amd-ctk-wrapper
```diff
@@ -0,0 +1,41 @@
...
+    echo "Error: Root path does not exist: $ROOT_PATH" >&2
+    exit 1
+fi
+
+if [ ! -x "$ROOT_PATH/usr/bin/amd-ctk" ]; then
+    echo "Error: amd-ctk not found at $ROOT_PATH/usr/bin/amd-ctk" >&2
+    exit 1
+fi
+
+exec chroot "$ROOT_PATH" /usr/bin/amd-ctk "$ARGS"
```
tomponline: More details here https://github.com/canonical/lxd/pull/17343#discussion_r2753627935

But this looks to be a problem to me, as it will always run `amd-ctk` after switching root to the host's filesystem, rendering the bundled `bin/amd-ctk` redundant. 

This will not be suitable for Ubuntu Core. 

Ideally we'd want to find a way to run the bundled `amd-ctk` command, but ask it to scan the host's filesystem. Like we do for `nvidia-ctk`

tomponline (changes requested):
https://github.com/canonical/lxd/pull/17343#pullrequestreview-3738576030

tomponline (approved):
Thanks!

tomponline:
Thanks @dmitry-lyfar 

FYI I've opened a follow up PR to pin the release version of amd-ctk and the Go version used to build it (like we do for other dependencies):

https://github.com/canonical/lxd-pkg-snap/pull/1036

PR #1022, opened 2026-01-12
qemu: Bump to import/1%10.2.0+ds-1

tomponline:
Related to https://github.com/canonical/lxd/issues/17361

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -785,7 +785,7 @@ parts:
       - libatomic
       - spice-server
     source: https://git.launchpad.net/ubuntu/+source/qemu
-    source-commit: e4d56a8af1739abbc298b92dc4d777d281ea278a # import/1%10.1.0+ds-5ubuntu4
+    source-commit: 261f8b1458c5341d3cb5a424d110ff5bc92d1b9d # import/1%10.2.0+ds-1
     source-depth: 1
     source-type: git
     plugin: autotools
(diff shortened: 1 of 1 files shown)


simondeziel (approved):

snapcraft.yaml
```diff
@@ -785,7 +785,7 @@ parts:
       - libatomic
       - spice-server
     source: https://git.launchpad.net/ubuntu/+source/qemu
-    source-commit: e4d56a8af1739abbc298b92dc4d777d281ea278a # import/1%10.1.0+ds-5ubuntu4
+    source-commit: 261f8b1458c5341d3cb5a424d110ff5bc92d1b9d # import/1%10.2.0+ds-1
```
simondeziel, line 788: Actually, this is not an Ubuntu branch so it's not even in Ubuntu Resolute yet.

simondeziel (changes requested):

snapcraft.yaml
simondeziel: Hmm, but maybe you intentionally went ahead of Ubuntu to address the crasher. I guess that's fair as long as we understand that we are ahead of the archive ;)

simondeziel (approved):

snapcraft.yaml
tomponline: thats right, but its coming soon to Resolute I'm told by @cpaelzer, so once tagged there we can switch to it.

PR #1030, opened 2026-01-30
qemu: update to 8.2.2+ds-0ubuntu1.12 (5.21-edge)

simondeziel:
Addresses https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/2131822 which is especially relevant to LXD due to using host CPU pass-through.

The corresponding .deb is still in `noble-proposed` but it cleared manual verification so it should land in `noble-updates` in a few days.


tomponline (approved):
ta

PR #1038, opened 2026-02-05
daemon.start: Link the pro command to the run-host wrapper (5.0-edge)

simondeziel:


simondeziel:
I mount binded the `daemon.start` and confirmed the `/run/bin/pro` symlink to be created at daemon start time.

tomponline (approved):

PR #1039, opened 2026-02-05
daemon.start: Link the pro command to the run-host wrapper (4.0-edge)

simondeziel:


simondeziel:
This should be safe to merge even if `stable-4.0` doesn't yet support looking pro attachment.

Note that I mount binded the `daemon.start` and confirmed the `/run/bin/pro` symlink to be created at daemon start time.

tomponline (approved):

PR #1040, opened 2026-02-05
Drop `ui.enable` config

simondeziel:
It makes very little sense to disable the LXD UI so let's have it always on[*].

*: except on `riscv64` where it is not available due to https://github.com/nodejs/snap/issues/77

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -51,7 +51,6 @@ description: |-
    - openvswitch.builtin: Run a snap-specific OVS daemon [default=false]
    - openvswitch.external: Use the system's OVS tools (ignores openvswitch.builtin) [default=false]
    - ovn.builtin: Use snap-specific OVN configuration [default=false]
-   - ui.enable: Enable the web interface [default=true]
    - zfs.external: Use the system's ZFS tools [default=false]
 
  For system-wide configuration of the CLI, place your configuration in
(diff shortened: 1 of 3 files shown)


simondeziel:
CC @edlerd

edlerd (commented):
Good idea!

I think we should also remove it from the docs:
* https://documentation.ubuntu.com/lxd/latest/howto/access_ui/#enable-or-disable-the-ui

Just today in a user interview, we had someone stumble on this, and it seems we can simply remove it by now.

simondeziel:
> I think we should also remove it from the docs:
> 
>     * https://documentation.ubuntu.com/lxd/latest/howto/access_ui/#enable-or-disable-the-ui

If this gets merged, I'll update the doc, thanks!

tomponline (commented):
@simondeziel what happens if a user who has set this key upgrades and the key isnt supported?

simondeziel:
> @simondeziel what happens if a user who has set this key upgrades and the key isnt supported?

The UI would be enabled silently.

tomponline (approved):

simondeziel:
> I think we should also remove it from the docs:
> 
>     * https://documentation.ubuntu.com/lxd/latest/howto/access_ui/#enable-or-disable-the-ui

https://github.com/canonical/lxd/pull/17579

PR #1047, opened 2026-02-10
Drop ZFS 2.1 (incompatible with Ubuntu 5.15 and earlier kernels)

simondeziel:
Ubuntu 24.04 GA kernel and 22.04 HWE kernel now become the minimum supported
kernels.

This means that 22.04 GA kernel (5.15) won't work anymore as it bundles ZFS 2.1
kmod.

This is in line with https://github.com/canonical/lxd/issues/16559

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1591,7 +1591,6 @@ parts:
       - xz
       - wrappers
       - xtables
-      - zfs-2-1
       - zfs-2-2
       - zfs-2-3
       - zstd
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1104,55 +1104,6 @@ parts:
       # Include the lzma symlink
       ln -s xz "${CRAFT_PART_INSTALL}/usr/bin/lzma"
 
-  zfs-2-1:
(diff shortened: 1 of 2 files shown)


tomponline:
@escabo is this ok to include in lxd 6.7 which will go into the latest/stable channel?

escabo:
> @escabo is this ok to include in lxd 6.7 which will go into the latest/stable channel?

@simondeziel tells me we have a workaround in the case where someone would need the old tooling so go for it.

simondeziel:
> > @escabo is this ok to include in lxd 6.7 which will go into the latest/stable channel?
> 
> @simondeziel tells me we have a workaround in the case where someone would need the old tooling so go for it.

Indeed, I mentioned `zfs.external` was a possible workaround for 22.04 users with the GA kernel desiring to run `latest/edge` as opposed to `5.21/stable`.

Hopefully, that population will be quite small and I think that `snap install lxd` defaulting to `5.21/stable` will have helped in making it as small as possible.

tomponline:
Thanks both

tomponline (approved):
Ta

PR #1067, opened 2026-02-19
qemu: fetch just enough history to allow restoring python wheels (5.0-edge)

simondeziel:
An alternative would be to do a full clone by dropping `source-depth: 1` but this results in more data being pulled from git:

```
$ du -sm qemu/ qemu+fetch/ qemu-full/
431	qemu/
461	qemu+fetch/
591	qemu-full/
```


simondeziel:
@tomponline this should fix the build failures on `5.0-edge`. I've reproduced the `git restore` issue and confirmed the partial fetch made it work to resurrect the python wheels.

tomponline (approved):

PR #1092, opened 2026-03-03
Build fixes (5.0-edge)

simondeziel:


simondeziel:
@tomponline this worked during a remote-build. However, it seems that GH is blocking or throttling git activity from LP builders as we are seeing widespread failures during cloning from various channels. I'll try pacing the retries slowly.

tomponline (approved):

PR #1096, opened 2026-03-04
Fix editor mount namespace blocking ZFS/LVM storage operations

### LINKED ISSUE #15664: Unable to delete instance while editing its profile
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for the bug I encountered.

### Distribution

Ubuntu

### Distribution version

22.04 (amd64)

### Output of "snap list --all lxd core20 core22 core24 snapd"

```shell
$ snap list --all lxd core20 core22 core24 snapd
Name    Version      Rev    Tracking       Publisher   Notes
core22  20250408     1963   latest/stable  canonical✓  base
core24  20250504     988    latest/stable  canonical✓  base
lxd     git-e208e86  33934  latest/edge    canonical✓  disabled,in-cohort
lxd     git-9ca5e0b  33951  latest/edge    canonical✓  in-cohort
snapd   2.68.4       24505  latest/stable  canonical✓  snapd
```

### Output of "lxc info" or system info if it fails

```shell
n/a
```

### Issue description

Deleting an instance fails if there is a live edit session of one of its profile.

### Steps to reproduce

1. Create instance
```
$ lxc launch ubuntu-minimal-daily:24.04 c1
```
2. Fire an idle edit on the `default` profile
```
$ lxc profile edit default &
[1] 981147
```
3. Try to delete the instance
```
$ lxc delete -f c1
Error: Failed deleting instance "c1" in project "default": Error deleting storage volume: Failed to run: zfs destroy -r default/containers/c1: exit status 1 (cannot destroy 'default/containers/c1': dataset is busy)

[1]+  Stopped                 lxc profile edit default

$ lxc delete -f c1
Error: Failed deleting instance "c1" in project "default": Error deleting storage volume: Failed to run: zfs destroy -r default/containers/c1: exit status 1 (cannot destroy 'default/containers/c1': dataset is busy)
```
4. Exit the idle editor
```
$ fg  # bring back the editor, quit it (`:q`) without making any change
```
5. Deleting the instance now works
```
$ lxc delete -f c1
```

### Information to attach

- [ ] Instance log (`lxc info NAME --show-log`)
- [ ] Instance configuration (`lxc config show NAME --expanded`)
- [ ] Main daemon log (at `/var/log/lxd/lxd.log` or `/var/snap/lxd/common/lxd/logs/lxd.log`)
- [ ] Output of the client with `--debug`
- [ ] Output of the daemon with `--debug` (or use `lxc monitor` while reproducing the issue)

tugbataluy:
## Problem
When running `lxc delete -f <instance>` while `lxc profile edit` has an editor open, ZFS and LVM containers fail to delete with "busy dataset/volume" error as mentioned in https://github.com/canonical/lxd/issues/15664.

## Root Cause
The editor wrapper script uses `unshare -m` to create a new mount namespace for the editor process. The system's mounts have shared propagation by default, so the new namespace inherits copies of all current mounts, including ZFS dataset and LVM logical volume mounts.
When `lxc delete -f` runs, LXD unmounts the storage in the main namespace, but the mount reference still exists in the editor's namespace. ZFS/LVM refuse to destroy the dataset because it's still mounted somewhere.

## Fix
Add `--propagation slave` to the `unshare` command. This ensures unmounts in the parent namespace propagate to the editor's namespace, preventing leftover mounts from blocking storage operations.

`--propagation private` was also considered but it doesn't work with ZFS block mode (volume.zfs.block_mode=true) the mount remains in the editor's namespace and blocks deletion. With slave mode, the unmount propagates from parent to child, cleanly removing the mount reference.

Fixes: https://github.com/canonical/lxd/issues/15664

### DIFF
--- a/snapcraft/wrappers/editor
+++ b/snapcraft/wrappers/editor
@@ -16,8 +16,11 @@ run_cmd() {
         fi
     fi
 
+    # Use --propagation slave so unmounts in the parent namespace propagate to
+    # the editor's namespace. This prevents leftover mounts (ZFS/LVM) from
+    # blocking storage delete operations while the editor is open.
     # shellcheck disable=SC2145
-    exec unshare --kill-child -U -m -p -r -f --root="/var/lib/snapd/hostfs/" "/bin/sh" -c "mount -t proc proc /proc 2>/dev/null || true; exec \"${CMD}\" ${IGNORERCFILES} \"$@\""
+    exec unshare --kill-child -U -m --propagation slave -p -r -f --root="/var/lib/snapd/hostfs/" "/bin/sh" -c "mount -t proc proc /proc 2>/dev/null || true; exec \"${CMD}\" ${IGNORERCFILES} \"$@\""
 }
 
 # Detect base name


tomponline:
Interesting I think you're on the right track here, thanks!

However:

> man unshared
--propagation private|shared|slave|unchanged
           Recursively set the mount propagation flag in the new mount
           namespace. **The default is to set the propagation to private**. It is
           possible to disable this feature with the argument unchanged. The
           option is silently ignored when the mount namespace (--mount) is
           not requested.


Is `private` not the default mode anyway?

Perhaps `unchanged` or `slave` would be more appropriate?

Did specifying `private` explicitly fix the issue?

tugbataluy:
> Interesting I think you're on the right track here, thanks!
> 
> However:
> 
> > man unshared
> > --propagation private|shared|slave|unchanged
> > Recursively set the mount propagation flag in the new mount
> > namespace. **The default is to set the propagation to private**. It is
> > possible to disable this feature with the argument unchanged. The
> > option is silently ignored when the mount namespace (--mount) is
> > not requested.
> 
> Is `private` not the default mode anyway?

According to man unshare, yes. But the system mounts have shared propagation by default:
```
$ cat /proc/self/mountinfo | grep shared:
31 1 8:1 / / rw,relatime shared:1 - ext4 /dev/sda1 ...
```
When `unshare -m` creates a new namespace, it inherits these shared mounts. The ZFS/LVM mounts get copied into the editor's namespace, and those references persist until the editor exits, blocking `zfs destroy` and `lvremove` even after the main namespace unmounts them.
Setting `--propagation private` explicitly ensures the editor's namespace doesn't hold references that block storage operations.

tomponline:
> > Interesting I think you're on the right track here, thanks!
> > However:
> > > man unshared
> > > --propagation private|shared|slave|unchanged
> > > Recursively set the mount propagation flag in the new mount
> > > namespace. **The default is to set the propagation to private**. It is
> > > possible to disable this feature with the argument unchanged. The
> > > option is silently ignored when the mount namespace (--mount) is
> > > not requested.
> > 
> > 
> > Is `private` not the default mode anyway?
> 
> According to man unshare, yes. But the system mounts have shared propagation by default:
> 
> ```
> $ cat /proc/self/mountinfo | grep shared:
> 31 1 8:1 / / rw,relatime shared:1 - ext4 /dev/sda1 ...
> ```
> 
> When `unshare -m` creates a new namespace, it inherits these shared mounts. The ZFS/LVM mounts get copied into the editor's namespace, and those references persist until the editor exits, blocking `zfs destroy` and `lvremove` even after the main namespace unmounts them. Setting `--propagation private` explicitly ensures the editor's namespace doesn't hold references that block storage operations.

So the man page is incorrect for unshare?

tugbataluy:
> > > Interesting I think you're on the right track here, thanks!
> > > However:
> > > > man unshared
> > > > --propagation private|shared|slave|unchanged
> > > > Recursively set the mount propagation flag in the new mount
> > > > namespace. **The default is to set the propagation to private**. It is
> > > > possible to disable this feature with the argument unchanged. The
> > > > option is silently ignored when the mount namespace (--mount) is
> > > > not requested.
> > > 
> > > 
> > > Is `private` not the default mode anyway?
> > 
> > 
> > According to man unshare, yes. But the system mounts have shared propagation by default:
> > ```
> > $ cat /proc/self/mountinfo | grep shared:
> > 31 1 8:1 / / rw,relatime shared:1 - ext4 /dev/sda1 ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When `unshare -m` creates a new namespace, it inherits these shared mounts. The ZFS/LVM mounts get copied into the editor's namespace, and those references persist until the editor exits, blocking `zfs destroy` and `lvremove` even after the main namespace unmounts them. Setting `--propagation private` explicitly ensures the editor's namespace doesn't hold references that block storage operations.
> 
> So the man page is incorrect for unshare?

No, not exactly:

Yes, the man page is correct `--propagation private` is the default for `unshare -m`. However, this only governs future mount/unmount propagation after namespace creation.

Inherited Mounts Retain Parent Propagation

When `unshare -m` runs, the new namespace inherits copies of all parent mounts first, including their propagation type. Systemd sets rootfs mounts to "shared" early in boot, so these copies maintain peer relationships across namespaces.

LVM/ZFS detect these shared references as "still mounted," blocking deletes even after umount. The default MS_REC|MS_PRIVATE recursion (util-linux ≥2.27) only privatizes future events, leaving inherited shared mounts active.

See:
https://stackoverflow.com/questions/56254904/why-can-a-process-launched-with-unshare-m-affect-the-mounts-in-the-host

tugbataluy:
## Why `--propagation slave` instead of `--propagation private`

### Mount Propagation Modes

- private: The namespace gets a copy of all mounts at creation time, but becomes completely isolated. Mount/unmount events in the parent do not propagate to the child.

- slave: The namespace receives mount/unmount events from the parent (one-way: parent -> child), but changes in the child don't affect the parent.

### Why slave is the right choice
With `--propagation private`, when LXD unmounts a storage volume in the main namespace, the editor's namespace still holds its copy of the mount. For ZFS block mode volumes, this leftover mount blocks deletion.

With `--propagation slave`, unmounts in the parent propagate to the editor's namespace. When LXD unmounts the storage, the editor's namespace receives that unmount event, cleanly releasing the mount reference.
The editor only needs to read/write a temporary YAML file , it has no need to create mounts or isolate itself from parent unmount events. One-way propagation (parent -> child) is exactly what we need.

tomponline:
> ## Why `--propagation slave` instead of `--propagation private`
> ### Mount Propagation Modes
> 
>     * private: The namespace gets a copy of all mounts at creation time, but becomes completely isolated. Mount/unmount events in the parent do not propagate to the child.
> 
>     * slave: The namespace receives mount/unmount events from the parent (one-way: parent -> child), but changes in the child don't affect the parent.
> 
> 
> ### Why slave is the right choice
> 
> With `--propagation private`, when LXD unmounts a storage volume in the main namespace, the editor's namespace still holds its copy of the mount. For ZFS block mode volumes, this leftover mount blocks deletion.
> 
> With `--propagation slave`, unmounts in the parent propagate to the editor's namespace. When LXD unmounts the storage, the editor's namespace receives that unmount event, cleanly releasing the mount reference. The editor only needs to read/write a temporary YAML file , it has no need to create mounts or isolate itself from parent unmount events. One-way propagation (parent -> child) is exactly what we need.

Sounds good!

tomponline:
@tugbataluy did you manage to prove my concern around private causing issues with releasing the block device btw?

also, are there any other places we call `unshare -m` in the snap packaging that are likely to have the same issue?

tugbataluy:
> @tugbataluy did you manage to prove my concern around private causing issues with releasing the block device btw?
> 
> also, are there any other places we call `unshare -m` in the snap packaging that are likely to have the same issue?

Hi Tom, yes I've managed to prove your concern about private propagation, that's why I decided to go with your suggestion.

<del>
And for the other occurrences of  ` unshare -m`:
run-host` and `virt-v2v-in-place` both implicitly create a new mount namespace via `--root` without `--propagation slave`. This means unmounts in the parent namespace won't propagate to those processes' namespaces.

As far as understand `virt-v2v-in-place` is particularly concerning:  VM conversion is a long-running operation, so it's exactly the kind of process that could hold a mount namespace open while a user tries to delete LXD storage, triggering "dataset is busy" errors.

run-host is lower risk since it's typically short-lived (single command execution), but the same issue could theoretically occur.

But what would you think about using propagation flags explicitly for those wrappers?
</del>

tomponline:
> But what would you think about using propagation flags explicitly for those wrappers?

Certainly seems like would be a good idea.

tugbataluy:
> > But what would you think about using propagation flags explicitly for those wrappers?
> 
> Certainly seems like would be a good idea.

Actually, looking again at those wrappers - they only use -U -r without -m:
`exec unshare -U -r --root="/var/lib/snapd/hostfs/" "${CMD}" "$@"`

Without `-m`, no new mount namespace is created, so `--propagation` isn't applicable. The mount namespace inheritance issue only occurs when -m is used.

So editor is the only wrapper affected by this bug.

Sorry for my confusion :disappointed:

tomponline (approved):
Thanks!

MggMuggins:
@tomponline Will this be backported to the stable snaps? I don't see any backports on the relevant branches.

tomponline:
> @tomponline Will this be backported to the stable snaps? I don't see any backports on the relevant branches.

yep it should be backportable indeed

tomponline:
> @tomponline Will this be backported to the stable snaps? I don't see any backports on the relevant branches.

- https://github.com/canonical/lxd-pkg-snap/pull/1185
- https://github.com/canonical/lxd-pkg-snap/pull/1186

MggMuggins:
Thank you!

PR #1109, opened 2026-03-10
lxd: build `lxc` statically and bypass confinement for completions

simondeziel:
When doing (inside my home):

```
$ cd ~
$ lxc file push <tab><tab>
...
```

Files in the current directory are listed as they should. However that is kind of by accident as the snap confinement is preventing the `lxc` command from doing the actual completion as we can see from this Apparmor denial:

```
audit: type=1400 audit(1772570126.835:3554): apparmor="DENIED" operation="open" class="file" profile="snap.lxd.lxc" name="/home/sdeziel/" pid=626160 comm="bash" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000
```

The [Go code](https://github.com/canonical/lxd/blob/main/lxc/completion.go#L1703-L1710) looks like this:

```go
	// Early return when no instances are found.
	if len(instances) == 0 {
		if includeLocalFiles {
			return nil, cobra.ShellCompDirectiveDefault
		}

		return instances, directives
	}
```

And `cobra.ShellCompDirectiveDefault == 0` which in turn is gracefully handled by the `lxc-completer.sh` script that has some fallback logic:

```shell
if (((directive & shellCompDirectiveNoFileComp) == 0)); then
      __lxc_debug "Listing files"
      _filedir
fi
```

Relying on the fallback logic is slower and cause log noise so instead, directly invoke the `lxc` command to bypass the confinement entirely and avoid the problem.

By statically compiling the `lxc` binary, it should be also be faster for all use as it will avoid the overhead of the dynamic linker and the CGO overhead.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1468,7 +1468,8 @@ parts:
       export CGO_LDFLAGS_ALLOW="(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"
 
       # Build the binaries
-      go build -trimpath -o "${CRAFT_PART_INSTALL}/bin/lxc" github.com/canonical/lxd/lxc
+      # lxc is built statically (CGO_ENABLED=0) so it can be executed unconfined on hosts with different libc during shell completion
+      CGO_ENABLED=0 go build -trimpath -o "${CRAFT_PART_INSTALL}/bin/lxc" -tags netgo github.com/canonical/lxd/lxc
 
       # Build LXD server binary into ${CRAFT_PART_INSTALL}/sbin/lxd so that it does not conflict with the
       # lxd-stophook wrapper script which is stored in ${CRAFT_PART_INSTALL}/bin/lxd.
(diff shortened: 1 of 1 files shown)


tomponline:
What does the fallback logic do? At what point is it failing due to external libc?

tomponline:
This will also result in not using the glibc resolver, which may have unintended impacts in the client resolving remote hosts in some setups.

tomponline:
https://pkg.go.dev/net#hdr-Name_Resolution

simondeziel:
> What does the fallback logic do?

The fallback logic is executed by the user's own shell `bash`, unconfined for real which is why it succeeds. The logic itself is the `_filedir` function as provided by `bash-completion`'s `/usr/share/bash-completion/bash_completion`.

> At what point is it failing due to external libc?

The LXD snap is built on `base: core24` which means one executing it on a host with a different glibc would run into issue due to bypassing the snap wrapper.

FYI, this PR is not yet ready for review, I'm still doing some tests to actually prove it fixes the problem it aims to.

simondeziel:
> This will also result in not using the glibc resolver, which may have unintended impacts in the client resolving remote hosts in some setups.

Yes indeed, `nsswitch.conf` would be ignored but `/etc/hosts` and `/etc/resolv.conf` would still be consulted.

simondeziel:
@kadinsayani you might be interested in this experiment. It might go nowhere but I think the problem on its own is interesting to look at.

kadinsayani:
> @kadinsayani you might be interested in this experiment. It might go nowhere but I think the problem on its own is interesting to look at.

Interesting findings, thanks for looping me in.

simondeziel:
So this works in avoiding the original Apparmor log spamming but it's only to be replaced by this:
```
Started snap.lxd.lxc-44c53a04-ced4-4e2f-943c-f1380f20bfd7.scope.
audit: type=1400 audit(1773369122.088:47380): apparmor="DENIED" operation="open" class="file" profile="snap.lxd.lxc" name="/proc/2647907/mountinfo" pid=2647907 comm="lxc" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000
```

Tab completion felt more responsive.

Experiment concluded: probably not worth going forward.

PR #1113, opened 2026-03-16
daemon.reload: wait for the signaled daemon PID to vanish

simondeziel:
The common idiom:

```
systemctl reload snap.lxd.daemon.service
lxd waitready --timeout=300
```

is racy due `systemctl reload` returning (immediately) before the new daemon is ready. This means that `lxd waitready` might start waiting on the old daemon that is shutting down.

### DIFF
--- a/snapcraft/commands/daemon.reload
+++ b/snapcraft/commands/daemon.reload
@@ -11,4 +11,19 @@ fi
 
 echo reload > "${SNAP_COMMON}/state"
 read -r PID < "${SNAP_COMMON}/lxd.pid"
-exec kill "$PID"
+kill "$PID"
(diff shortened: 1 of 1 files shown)


simondeziel:
Converted to draft until further discussion.

simondeziel:
@tomponline would you like to have this merged as is and I add the HUP rework in a future PR when the LXD side is ready?
I'm asking because it addresses a legitimate race that we sometimes hit for the 2 scripts that try to use LVM external tooling, irrespective of the `dnsmasq` killing.

tomponline:
I'll hold off for the sigup stuff i think, ta

PR #1120, opened 2026-03-22
edk2: ship secureboot and non-secureboot firmware variants

simondeziel:
Previously, the snap only shipped the secureboot CODE variant
(`SMM_REQUIRE=TRUE` on x86_64), which was used for all VMs regardless of their
SB setting. Being built with `SMM_REQUIRE=TRUE` meant that all UEFI variable
access (reads and writes to NVRAM) goes through System Management Mode, which
forces every vCPU into a synchronous rendezvous. This happens even when no
Secure Boot keys are enrolled. Incidentally, when `boot.mode=uefi-nosecureboot`
was used, the SB CODE was simply paired with empty VARS which puts the firmware
into "Setup Mode" (SB-capable, keys not yet enrolled). The entire Secure Boot
infrastructure is initialized: signature databases, certificate stores, the
SecureBootEnable variable, etc.

A non-SB CODE genuinely doesn't have Secure Boot support compiled in so there's
no Setup Mode ambiguity, and guest OS/bootloaders that check SecureBoot EFI
variables see a clear "not supported" rather than "supported but not
configured."

The solution is to ship both secureboot (`.ms`) and non-secureboot CODE
firmware files which will allow LXD to avoid the SMM overhead for VMs that
don't need it (`boot.mode=uefi-nosecureboot`).

This is done by switching build targets from the monolithic
`install-ovmf`/`install-aarch64` to the split `install-*-secboot` +
`install-*-no-secboot` targets available in edk2 2025.11.

At the same time, adopt names from the Ubuntu edk2 package (`OVMF_CODE_4M`/
`AAVMF_CODE` on the non-SB side, `OVMF_CODE_4M.ms`/`AAVMF_CODE.ms` on the SB
side). This also uncovered that LXD uses `OVMF` even on `arm64` where it should
instead be `AAMF`.

Until the LXD side can be updated to use different CODE for SB and non-SB and
also use the proper `AAVMF` name, backward-compat symlinks are added. This is
just a temporary situation and this will be removed once current LXD is
updated.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1408,7 +1408,8 @@ parts:
       - lib/python3/dist-packages/pyuefivars*
 
   lxd:
-    source: https://github.com/canonical/lxd
+    source: https://github.com/simondeziel/lxd
+    source-branch: edk2-code-vars-files
     # XXX: We often cherry-pick for candidate builds so don't do shallow clone (0 means full clone).
     source-depth: 0
     source-type: git
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -477,52 +477,6 @@ parts:
       # Build Debug
       run_build DEBUG ".debug"
 
-      # Firmware files summary
-      #
-      # The snap ships the following real files per architecture:
-      #
-      #   x86_64:
-      #     File                     | Role               | Used by
-      #     OVMF_CODE_4M.ms.fd       | SB CODE            | new LXD (secureboot)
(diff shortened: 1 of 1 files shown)


simondeziel:
I've only tested on amd64 and arm64 without the corresponding LXD changes. No regression but also no improvement from `PV_VARS` on amd64 with SB enabled.

On amd64, debug boot still shows that slow boot every 30-something reboots accompanied with FTW reclaims.
On arm64, still impossible to use debug boot but that's normal and expected (not a regression).

PR #1130, opened 2026-03-25
Switch to `core24` (5.21-edge)

simondeziel:


simondeziel:
## `amd64` smoke test

```
root@v3:~# snap list 
Name    Version        Rev    Tracking       Publisher   Notes
core24  20260211       1499   latest/stable  canonical✓  base
lxd     0+git.79ab996  x1     latest/edge    -           -
snapd   2.74.1         26382  latest/stable  canonical✓  snapd
root@v3:~# lxc shell v1
root@v1:~# dmidecode | grep -A1 -i edk
	Vendor: Ubuntu distribution of EDK II
	Version: 2024.02-2ubuntu0.7
```

## `arm64` smoke test

```
root@ubuntu:~# snap list
Name    Version        Rev    Tracking       Publisher   Notes
core24  20260211       1500   latest/stable  canonical✓  base
lxd     0+git.79ab996  x1     latest/edge    -           -
snapd   2.74.1         26383  latest/stable  canonical✓  snapd
root@ubuntu:~# lxc shell v1
root@v1:~# dmidecode | grep -A1 -i edk
	Vendor: Ubuntu distribution of EDK II
	Version: 2024.02-2ubuntu0.7
root@v1:~# uname -a
Linux v1 6.8.0-106-generic #106-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar  6 07:44:40 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux
```

snapcraft/commands/lxd
```diff
@@ -22,8 +22,8 @@ export PATH="${PATH}:${SNAP_CURRENT}/bin"
 export LXD_DIR="${LXD_DIR:-"${SNAP_COMMON}/lxd/"}"
 
 # Make sure we have a ZFS binary on the path
-export LD_LIBRARY_PATH="${SNAP_CURRENT}/zfs-2.1/lib/:${LD_LIBRARY_PATH}"
-export PATH="${SNAP_CURRENT}/zfs-2.1/bin:${PATH}"
+export LD_LIBRARY_PATH="${SNAP_CURRENT}/zfs-2.2/lib/:${LD_LIBRARY_PATH}"
```
tomponline, line 25: we cant change the min ZFS version in an LTS

snapcraft.yaml
```diff
@@ -1434,13 +1434,23 @@ parts:
   uefivars:
     after:
       - spice-server
-      - qemu-ovmf-secureboot
+      - edk2
       - nftables
+    # XXX: cannot use python3-pyuefivars (1.0.0-0ubuntu2) from Noble as the Ubuntu patch breaks it
```
tomponline, line 1439: Can we use 26.04 sources so we get LTS support (for both 6 and 5.21)?

snapcraft/commands/daemon.start
```diff
@@ -39,6 +39,13 @@ fi
 # Detect base name
 SNAP_BASE="$(sed -n '/^name:/ s/^name:\s*\(core[0-9]\{2\}\)/\1/p' /meta/snap.yaml)"
 
+# Temporary hack to workaround systemctl reload snap.lxd.daemon
+# problem with core24-based LXD snap
+_LXD_SNAP_DEVCGROUP_CONFIG="/var/lib/snapd/hostfs/var/lib/snapd/cgroup/snap.lxd.device"
```
tomponline, line 44: do we need this anymore in 6 or 5.21? I'd prefer to not include it if not needed.

tomponline (approved):
Thanks! LGTM

Just a couple of follow up questions/potential changes.

snapcraft/commands/lxd
simondeziel: That's not changing the min ZFS version. It seems it was just to have a working `zfs` and `zpool` command when one would use `snap run lxd` which runs LXD in the "foreground" instead of as a daemon.

Rather than hardcode a specific version, I'll extract the version detection, see https://github.com/canonical/lxd-pkg-snap/pull/1132

snapcraft/commands/daemon.start
simondeziel: Let's try once more: https://github.com/canonical/lxd-pkg-snap/pull/1133
This time, stars seem aligned ;)

snapcraft.yaml
simondeziel: For 6, since it will move to `core26` this will become a "stage-package" part so I'd rather not consume sources for it.

You make a case for 5.21 that will stay on `core24` though.

PR #1132, opened 2026-03-26
Improve ZFS version/PATH handling for `snap run lxd` use case

simondeziel:


simondeziel:
I've confirmed that `daemon.start` keeps picking the right ZFS version and that now `snap run lxd` too while it previously picked 2.2 unconditionally.

tomponline (approved):

PR #1159, opened 2026-04-13
Source `uefivars` from 26.04 sources on Launchpad

simondeziel:


simondeziel:
This was successfully tested with `tests/efi-vars-editor-vm` (the NVRAM toture version at that).

tomponline (approved):

PR #1162, opened 2026-04-14
edk2: use the hash tied to 2025.11-3ubuntu7 

simondeziel:
Also ask Copilot to watch for such easy mistakes.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,14 @@
+# Copilot instructions for lxd-pkg-snap
+
+## snapcraft.yaml: source-commit entries
+
+Every `source-commit:` line has the form:
+
+```
+source-commit: <sha1> # <version-tag>
+```
+
+**Always update both fields together.** When bumping a dependency to a new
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,14 @@
...
+Every `source-commit:` line has the form:
+
+```
+source-commit: <sha1> # <version-tag>
+```
+
+**Always update both fields together.** When bumping a dependency to a new
+version, the SHA1 hash *and* the version comment must change at the same time.
+Updating only the comment (leaving the old SHA1) silently pins the build to the
+wrong upstream revision.
```
simondeziel, line 14: This was generated by Claude when asked to have AI watch our backs.

tomponline (approved):
Whoops! Thanks

simondeziel:
> Whoops! Thanks

FYI, I'm augmenting `lxd-snapcraft` with a `-verify-source-commit` that will be wired to CI so we never run into this again. I'm also doing it as part of https://github.com/canonical/lxd/issues/16544

PR #1163, opened 2026-04-14
Snap tooling

### LINKED ISSUE #16544: `actions/lp-snap-build` should be moved to `lxd-ci.git`
### Please confirm

- [x] I have searched existing issues to check if an issue already exists for my feature request.

### Is your feature request related to a problem? Please describe.

The `actions/lp-snap-build` already depends on the `lxd-snapcraft` tool that's coming from `lxd-ci.git` so it would make sense to move the action code there too.

### Describe the solution you'd like

Move the action code to `lxd-ci.git`.

### Describe alternatives you've considered

Status quo

### Additional context

_No response_

simondeziel:
Fixes https://github.com/canonical/lxd/issues/16544

### DIFF
--- a/.github/actions/lp-snap-build/action.yml
+++ b/.github/actions/lp-snap-build/action.yml
@@ -1,0 +1,69 @@
+name: Setup SSH for integration with snap build on Launchpad
+description: |
+  This action sets up SSH access to Launchpad for pushing and pulling code and signing commits.
+  It also configures Git with the necessary settings for signing commits.
+  It clones the snap packaging repository into ~/${PACKAGE}-pkg-snap-lp
+  Lastly it installs the `lxd-snapcraft` tool to allow modifying the snapcraft.yaml file.
+
+  It relies on some environment variables being set:
+  - `SSH_AUTH_SOCK`: The path to the SSH agent socket.
+  - `REPO`: The URL of the snap packaging repository.
+  - `BRANCH`: The branch to clone from the snap packaging repository.
+  - `PACKAGE`: The name of the snap package.
+
+inputs:
+  ssh-key:
+    description: The private SSH key for pulling/pushing to Launchpad and commit signing.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Configure Git
+      shell: bash
+      run: |
+        set -eux
+        git config --global transfer.fsckobjects true
+        git config --global user.name "Canonical LXD Bot"
+        git config --global user.email "lxd@lists.canonical.com"
+        git config --global commit.gpgsign true
+        git config --global gpg.format "ssh"
+        git config --global user.signingkey ~/.ssh/id_ed25519.pub
+
+    - name: Install lxd-snapcraft tool
+      shell: bash
+      run: |
+        # Depends on Go being available which is the case in the GitHub Actions environment
+        set -eux
+        cd "${{ github.action_path }}/lxd-snapcraft"
+        go build -o /tmp/lxd-snapcraft .
+        sudo mv /tmp/lxd-snapcraft /usr/local/bin/
+
+    - name: Verify Launchpad SSH host key
+      shell: bash
+      run: |
+        set -eux
+        mkdir -m 0700 -p ~/.ssh/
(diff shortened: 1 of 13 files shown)


simondeziel:
@tomponline this new `-verify-source-commits` flag that's now used during CI tests. It was also confirmed to catch the edk2 version bump issue after a test revert.

.github/actions/lp-snap-build/action.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    - name: Verify Launchpad SSH host key
+      shell: bash
+      run: |
+        set -eux
+        mkdir -m 0700 -p ~/.ssh/
+        # In ephemeral environments like GitHub Action runners, relying on TOFU isn't providing any security
+        # so require the key obtained by `ssh-keyscan` to match the expected hash from https://help.launchpad.net/SSHFingerprints
+        ssh-keyscan git.launchpad.net >> ~/.ssh/known_hosts
+        [ "$(ssh-keygen -qlF git.launchpad.net)" = 'git.launchpad.net RSA SHA256:UNOzlP66WpDuEo34Wgs8mewypV0UzqHLsIFoqwe8dYo' ]
```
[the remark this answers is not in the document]
simondeziel: This has worked fine as is for a while, let's leave it like that.

.github/actions/lp-snap-build/lxd-snapcraft/lxd-snapcraft_test.go
```diff
@@ -0,0 +1,207 @@
...
+func TestVerifySourceCommits_RealFile(t *testing.T) {
+	config, err := loadSnapcraftYaml(repoSnapcraftYaml)
+	if err != nil {
+		t.Fatalf("loadSnapcraftYaml failed: %v", err)
+	}
+
+	err = verifySourceCommits(repoSnapcraftYaml, config)
+	if err != nil {
+		t.Fatalf("verifySourceCommits failed on repo snapcraft.yaml: %v", err)
+	}
```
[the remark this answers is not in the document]
simondeziel: It's actually the intended behavior.

.github/actions/lp-snap-build/lxd-snapcraft/lxd-snapcraft.go
```diff
@@ -25,37 +30,47 @@ func main() {
 		log.Fatal(err)
 	}
 
-	if *flagPackageName == "" {
-		log.Fatal("Package name is required")
+	if *flagVerifySourceCommits {
+		err = verifySourceCommits(*flagFilePath, snapcraftConfig)
```
tomponline: @simondeziel given that this function is called from the unit tests to test the actual snapcraft.yaml (which caught me by surprised to be honest), I noticed the tests were creating a copy of the snapcraft.yaml file and then passing that file path to verifySourceCommits.

Given the size of the file I would encourage you to update this function to accept an io.Reader and then pass a file handle to to it instead.

This way the actual implementation and the unit tests won't need to make a copy of the file (and delete it etc).

simondeziel: > @simondeziel given that this function is called from the unit tests to test the actual snapcraft.yaml (which caught me by surprised to be honest),

I agree that's unusual but I did it that way to avoid needing an extra job/VM to run the tests.

> I noticed the tests were creating a copy of the snapcraft.yaml file and then passing that file path to verifySourceCommits.
> 
> Given the size of the file I would encourage you to update this function to accept an io.Reader and then pass a file handle to to it instead.
> 
> This way the actual implementation and the unit tests won't need to make a copy of the file (and delete it etc).

Ack, will do.

tomponline: Could it be invoked directly as a separate step within the same workflow?

simondeziel: Cleaner thanks, done.

tomponline (approved):
Nice thanks!

PR #1188, opened 2026-04-22
lxd: suppress errors from sourcing `setup-zfs` helper

simondeziel:
This should avoid console pollution:

```
==> Unsupported ZFS version (2.1)
Consider installing ZFS tools in the host and use zfs.external
```

And `daemon.start` that also uses that helper should cause the same message to go into the journal.


tomponline (approved):
ta

Please can you backport to latest-candidate too ta

PR #1196, opened 2026-04-23
qemu: remove unneeded `pycotap` wheel

simondeziel:
Debian carries a patch that disables `pycotap` (only used for tests) since version 9.2.0+ds-5

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -919,12 +919,11 @@ parts:
 
       set -eux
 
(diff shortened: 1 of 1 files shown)


simondeziel:
@tomponline I know I told you I wouldn't submit this one until the switch to `core26` but since it will be delayed, I'd rather have this merged now.

tomponline (approved):

PR #1216, opened 2026-05-22
RISC-V: enable virtual machines

xypron:
This series contains all that is needed to launch virtual machines on RISC-V with LXD. Additionally ceph, ovn, zfs are enabled for RISC-V.

OpenSBI and libatomic1 need to be staged as additional libraries.
QEMU cannot be built without --enable-tcg

With Linux kernel v7.0 booting with ACPI did not work correctly. This is correct with the accompanying pull-request https://github.com/canonical/lxd/pull/18318 . Cf. https://bugs.launchpad.net/ubuntu/+source/linux-riscv/+bug/2153582.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -933,6 +933,7 @@ parts:
         - libusb-1.0-0
         - libusbredirparser1t64
         - liburing2
+        - opensbi # This is needed due to --disable-install-blobs.
         - seabios # This is needed due to --disable-install-blobs.
         - qemu-system-data # This is needed due to --disable-install-blobs.
     override-stage: |-
@@ -994,6 +995,7 @@ parts:
       usr/local/libexec/: bin/
       usr/local/share/: share/
       usr/share/qemu/kvmvapic.bin: share/qemu/
+      usr/share/qemu/opensbi*: share/qemu/
       usr/share/qemu/s390-ccw.img: share/qemu/
       usr/share/qemu/slof.bin: share/qemu/
       usr/share/seabios/bios-256k.bin: share/qemu/
@@ -1014,6 +1016,7 @@ parts:
       - share/qemu/bios-256k.bin
       - share/qemu/efi-virtio.rom
       - share/qemu/kvmvapic.bin
+      - share/qemu/opensbi-riscv64-generic-fw_dynamic.bin
       - share/qemu/s390-ccw.img
       - share/qemu/slof.bin
       - share/qemu/vgabios-bochs-display.bin
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -241,19 +241,15 @@ parts:
       usr/lib/: lib/
     override-stage: |-
       [ "$(uname -m)" = "armv7l" ] && exit 0
-      [ "$(uname -m)" = "riscv64" ] && exit 0
       craftctl default
     override-prime: |-
       [ "$(uname -m)" = "armv7l" ] && exit 0
-      [ "$(uname -m)" = "riscv64" ] && exit 0
       craftctl default
     override-pull: |-
       [ "$(uname -m)" = "armv7l" ] && exit 0
-      [ "$(uname -m)" = "riscv64" ] && exit 0
       craftctl default
     override-build: |-
       [ "$(uname -m)" = "armv7l" ] && exit 0
-      [ "$(uname -m)" = "riscv64" ] && exit 0
       craftctl default
     prime:
       - bin/ceph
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
tomponline, line 1: Please can you update the comment in this change to link to the upstream bug?

snapcraft.yaml
```diff
@@ -1014,6 +1016,7 @@ parts:
       - share/qemu/bios-256k.bin
       - share/qemu/efi-virtio.rom
       - share/qemu/kvmvapic.bin
+      - share/qemu/opensbi-riscv64-generic-fw_dynamic.bin*
```
simondeziel: Looking at https://packages.ubuntu.com/resolute/riscv64/qemu-system-riscv/filelist, I'm wondering why there is this trailing `*`?

snapcraft.yaml
```diff
@@ -452,6 +460,15 @@ parts:
           SRC_VARS_MS="AAVMF_VARS.ms.fd"
           # See note in `x86_64` section for DEBUG_VERBOSE explanation
           SED_EXPR="s#DEBUG_PRINT_ERROR_LEVEL = 0x8000004F#DEBUG_PRINT_ERROR_LEVEL = 0x8040004F#g"
+      elif [ "$(uname -m)" = "riscv64" ]; then
+          DSC="OvmfPkg/RiscVVirt/RiscVVirtQemu.dsc"
+          BUILD_TARGET="install-qemu-efi-riscv64"
+          SRC_DIR="debian/tmp/usr/share/qemu-efi-riscv64"
+          SRC_CODE="RISCV_VIRT_CODE.fd"
```
simondeziel, line 467: There is no `.secboot.fd` suffix so I'm wondering if SecureBoot is a thing on RISC-V?

snapcraft.yaml
```diff
@@ -390,17 +386,25 @@ parts:
...
         - uuid-dev
         - xorriso
+      - to riscv64:
+        - acpica-tools
+        - g++
+        - lsb-release
+        - nasm
+        - quilt
+        - rsync
+        - uuid-dev
```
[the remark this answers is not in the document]
xypron: EDK 2 builds fine with the selected boot dependency. Why should we add unused tools?

snapcraft.yaml
xypron: Shim does not exist for RISC-V yet.
There is no signing party for secure boot (like Microsoft for x86 and ARM).
We don't have signed kernels.

snapcraft.yaml
xypron: The wildcard csn be removed.

snapcraft.yaml
```diff
@@ -974,6 +974,13 @@ parts:
       sed -i "s/^unset target_list$/target_list=\"${QEMUARCH}-softmmu\"/" configure
       sed -i 's#libseccomp_minver=".*#libseccomp_minver="0.0"#g' configure
 
+      # As of May 2026 QEMU for riscv64 does not support building with --disable-tcg
+      if [ "${QEMUARCH}" = "riscv64" ]; then
+          # `set -- "$@"`: preserve arguments passed to `configure` and append
+          # `--enable-tcg` at the end to override the previous `--disable-tcg`.
+          sed -i '1 a set -- "$@" --enable-tcg' configure
```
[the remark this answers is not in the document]
xypron: Is the a notation to make single configuration flags architecture dependent?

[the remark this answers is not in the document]
simondeziel: This inline `sed` is fine by me.

snapcraft.yaml
simondeziel: Let's remove then, please. This way, if the file is missing, it will fail loudly.

snapcraft.yaml
xypron: I have added the link to the Launchpad bug where we investigate the issue.

snapcraft.yaml
xypron: I have updated the patch "RISC-V: stage OpenSBI".

simondeziel (approved):

simondeziel:
@xypron As I'm working on switching to `core26`, the `qemu-system-data` package will be pulled from Resolute instead of Noble. This runs into a missing `share/qemu/opensbi-riscv64-generic-fw_dynamic.bin` problem.

https://packages.ubuntu.com/noble-updates/all/qemu-system-data/filelist => has it
https://packages.ubuntu.com/resolute/all/qemu-system-data/filelist => doesn't have it

It seems to have moved to https://packages.ubuntu.com/resolute/riscv64/qemu-system-riscv/filelist
This leaves me wondering why `opensbi` was added to the `stage-packages`.

xypron:
@simondezie
In current Debian/Ubuntu packaging QEMU only provides a symbolic link to the OpenSBI firmware. The firmware itself is provided by the opensbi package. That symbolic link may have been moved as it is only relevant for RISC-V.

(In upstream QEMU OpenSBI is a git submodule and OpenSBI is built and provided as part of QEMU.)

simondeziel:
> @simondezie In current Debian/Ubuntu packaging QEMU only provides a symbolic link to the OpenSBI firmware. The firmware itself is provided by the opensbi package. That symbolic link may have been moved as it is only relevant for RISC-V.

Lovely, thanks!

PR #1221, opened 2026-05-27
github: add smoke-test to tests workflow

simondeziel:


.github/workflows/tests.yml
```diff
@@ -73,10 +75,74 @@ jobs:
...
+
+  smoke-test:
+    name: Smoke test snap
+    needs: build
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Download snap artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          artifact-ids: ${{ needs.build.outputs.snap-artifact-id }}
```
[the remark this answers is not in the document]
simondeziel: Using `name: lxd-snap` does not work with `archive: false` which I'd like to keep using. Also, it seems to work just fine as-is.

.github/workflows/tests.yml
```diff
@@ -73,10 +75,74 @@ jobs:
...
+
+          # Container instance
+          lxc launch ubuntu-minimal-daily:24.04 c1
+
+          # Allow for boot to complete
+          sleep 30
+
+          lxc list
+          lxc exec v1 -- hostname
+          lxc exec c1 -- hostname
```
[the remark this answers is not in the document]
simondeziel: The `sleep 30` should be enough, especially since the `lxc launch` of the container won't be instantaneous as it needs to pull the image from the network and extract it. Previous tests showed this `lxc launch` would take ~7s.

`lxc exec  --wait` is not a thing unfortunately.

.github/workflows/tests.yml
```diff
@@ -73,10 +75,74 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Download snap artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          artifact-ids: ${{ needs.build.outputs.snap-artifact-id }}
+
+      # An official snap from the store is needed to allow `--dangerous` afterward
+      - name: Setup LXD
+        uses: canonical/setup-lxd@a517a5346a08c624d81953578813a4fc0fe46e0b # main
```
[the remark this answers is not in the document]
simondeziel: Tags/releases are not frequently published for `canonical/setup-lxd`. Fortunately, this is also a very slow moving repo.

simondeziel:
@tomponline this one might be interesting to land before #1222 as it surfaces interesting data from the built snap.

.github/workflows/tests.yml
```diff
@@ -80,3 +80,74 @@ jobs:
...
+      - name: Download snap artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: lxd-snap
+
+      # An official snap from the store is needed to allow `--dangerous` afterward
+      - name: Setup LXD
+        uses: canonical/setup-lxd@a517a5346a08c624d81953578813a4fc0fe46e0b # main
+        with:
+          channel: latest/edge
```
tomponline: should this use the appropriate channel for the branch?

.github/workflows/tests.yml
```diff
@@ -80,3 +80,74 @@ jobs:
...
+
+  smoke-test:
+    name: Smoke test snap
+    needs: build
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Download snap artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: lxd-snap
```
tomponline: what does this do?

.github/workflows/tests.yml
```diff
@@ -70,13 +72,85 @@ jobs:
...
+        id: upload-snap
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: lxd-snap
           path: ${{ steps.snapcraft.outputs.snap }}
           if-no-files-found: error
           retention-days: 1
           archive: false
+
+  smoke-test:
```
tomponline: do we need to do this inside another runner, and download the artefact we just created? Or can we use the build runner?

simondeziel: Not strictly needed indeed. I'll rework to use the build runner to also be where the smoke test is done.

The 2 runners approach would make sense if we wanted to do smoke tests on a matrix like `ubunut-22.04`, `ubuntu-24.04` and eventually `ubuntu-26.04`...

But yeah, let's start simple ;)

.github/workflows/tests.yml
simondeziel: It doesn't matter at the functional level but indeed we could be smart and be sure to pick a snap branch that bring the same core as what we are about to smoke-test.

Let me have a think.

.github/workflows/tests.yml
simondeziel: I've drop it for now but it was just downloading the freshly built snap from the other runner.

simondeziel:
@tomponline I've addressed your feedback, please take another look when you can. Thanks!

tomponline (approved):
ta

PR #1224, opened 2026-05-29
qemu: avoid dependency on `pip` to download `meson` wheel

simondeziel:
The ESM update for `python3-pip`
(https://ubuntu.com/security/notices/USN-8344-1) introduced a regression: https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/2154576

```
:: + set -eux
:: + pip download --dest python/wheels meson==1.9.0
:: ERROR: Exception:
...
:: File "/usr/lib/python3.12/json/decoder.py", line 353, in raw_decode
:: obj, end = self.scan_once(s, idx)
:: ^^^^^^^^^^^^^^^^^^^^^^
:: json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 40961 (char 40960)
```

Since we are using `pip` just for downloading a single wheel, it should be simple enough to use `curl` instead.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -911,6 +911,7 @@ parts:
       - else:
         - bison
         - bzip2
+        - curl
         - flex
         - pkg-config
         - libaio-dev
@@ -923,7 +924,6 @@ parts:
         - liburing-dev
         - libusb-1.0-0-dev
         - libusbredirparser-dev
-        - python3-pip
         - quilt
         - librbd-dev
     stage-packages:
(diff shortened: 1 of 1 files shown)


simondeziel:
The faulty ESM package was fixed by another upload where the problematic patch was dropped (temporarily): https://ubuntu.com/security/notices/USN-8344-2

I still think it's a good idea to avoid this extra dependency as we only need to download a static file.

snapcraft.yaml
```diff
@@ -954,10 +954,13 @@ parts:
       set -eux
 
       # Download meson Python wheel (normally shipped in upstream tarballs)
-      pip download --dest python/wheels meson==1.9.0
+      # Equivalent to: pip download --dest python/wheels meson==1.9.0
+      mkdir -p python/wheels
+      MESON="meson-1.9.0-py3-none-any.whl"
+      curl --fail -sSL --output "python/wheels/${MESON}" "https://files.pythonhosted.org/packages/23/ed/a449e8fb5764a7f6df6e887a2d350001deca17efd6ecd5251d2fb6202009/${MESON}"
```
tomponline, line 961: Need to include in comment how to get url to update this in the future please.

simondeziel:
@tomponline `latest-edge` no longer needs but let me know if you'd like it in older branches or not.

PR #1226, opened 2026-05-29
SECURITY: new policy

simondeziel:
Encourage reporters to open advisories on LXD's main repo even if they are snap specific.


morphis (approved):
LGTM

PR #1227, opened 2026-05-30
edk2: delay boot only when `boot.debug_edk2=true`

simondeziel:


simondeziel:
The shorter boot was timed with a `boot.debug_edk2=true` VM and:

```
# cat firmware-timing 
#!/bin/bash
set -eu

INSTANCE="v1"
LOG_PATH="/var/snap/lxd/common/lxd/logs/${INSTANCE}/edk2.log"

# 1. Force the VM to stop
lxc stop -f "$INSTANCE" >/dev/null 2>&1
# Wipe the previous log history so we start clean
true > "$LOG_PATH"

# 2. Start a background worker that monitors the log file.
# It records the precise host timestamp the millisecond QEMU spawns,
# and records the timestamp the millisecond '[Bds]Booting' appears.
(
    # Wait for the log file to be recreated by QEMU
    while [ ! -s "$LOG_PATH" ]; do sleep 0.01; done
    START_TIME=$(date +%s.%N)

    # Tail the log until we hit the boot block
    tail -f -n +1 "$LOG_PATH" | grep -m 1 "Booting" >/dev/null
    END_TIME=$(date +%s.%N)

    # Compute high-resolution difference
    DURATION=$(echo "$END_TIME - $START_TIME" | bc)
    echo "----------------------------------------"
    echo "Firmware Execution Duration: ${DURATION}s"
    echo "----------------------------------------"
) &

# 3. Fire up the instance
lxc start "$INSTANCE" >/dev/null
wait
```

From this PR:
```
# bash ./firmware-timing 
----------------------------------------
Firmware Execution Duration: 30.831689846s
----------------------------------------
```

From `latest/edge`:

```
# bash ./firmware-timing 
----------------------------------------
Firmware Execution Duration: 38.170675808s
----------------------------------------
```

patches/edk2-0003-boot-delay.patch
```diff
@@ -1,29 +0,0 @@
...
---- a/OvmfPkg/OvmfPkgX64.dsc
-+++ b/OvmfPkg/OvmfPkgX64.dsc
-@@ -597,7 +597,7 @@
-   gUefiOvmfPkgTokenSpaceGuid.PcdPciMmio64Size|0x800000000
- !endif
- 
--  gEfiMdePkgTokenSpaceGuid.PcdPlatformBootTimeOut|0
-+  gEfiMdePkgTokenSpaceGuid.PcdPlatformBootTimeOut|3
- 
-   # Set video resolution for text setup.
```
tomponline, line 25: What about this bit? It doesnt seem to be timeout related?

snapcraft.yaml
```diff
@@ -484,15 +483,18 @@ parts:
           if [ "${TYPE}" = "DEBUG" ]; then
               # Add `DEBUG_VERBOSE` to print error level for DEBUG build
               sed --in-place=.bak "${SED_EXPR}" "${DSC}"
+              # Bump boot timeout to 3s for DEBUG builds
+              sed -i 's/\(gEfiMdePkgTokenSpaceGuid\.PcdPlatformBootTimeOut\)|[0-9]\+$/\1|3/' "${DSC}"
```
tomponline, line 487: I wonder if it would be more correct to use patch or git to apply a patch file rather thsn sed, so that in the future we will get a patch failure if it fails to apply?

patches/edk2-0003-boot-delay.patch
simondeziel: If you are wondering about the video resolution bit, it is just patch "context".

snapcraft.yaml
simondeziel: I went with `sed` for ease of reversibility and also to be idempotent if done after a failed `snapcraft pack`.

simondeziel:
This will need to be worked out by using patches but this will happen in the main LXD repo now that they are merged for `latest/edge`.

PR #1242, opened 2026-06-11
`core26` workarounds

simondeziel:
This should fix the following tests from `lxd-ci`:

* `csi`
* `storage-metadata`


simondeziel:
@tomponline I've bind mounted the relevant files to confirm they indeed fix the issues.

tomponline (approved):
Nice one thanks

PR #1245, opened 2026-06-12
zfs updates

simondeziel:


simondeziel:
I tested all 3 ZFS versions in local VMs.

tomponline (approved):

PR #1258, opened 2026-06-16
Dependency updates (latest-edge)

tomponline:
https://warthogs.atlassian.net/browse/LXD-4550

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -230,7 +230,7 @@ parts:
 
   amd-ctk:
     source: https://github.com/ROCm/container-toolkit
-    source-commit: 3ab0394f71220e693ae016fa6581fbd04a51ed44 # v1.2.0
+    source-commit: 59f21d9fdeb4a9373607b90352dbda272c721979 # v1.3.0
     source-depth: 1
     source-type: git
     plugin: go
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -644,12 +644,12 @@ parts:
 
   nvidia-container:
     source: https://github.com/NVIDIA/libnvidia-container
-    source-commit: 7585946c6471402577e14474d7c56ca5be0348d7 # v1.19.0
+    source-commit: 7585946c6471402577e14474d7c56ca5be0348d7 # v1.19.1
     source-depth: 1
     source-type: git
     plugin: make
     build-environment:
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -370,7 +370,7 @@ parts:
       - qemu
     source: https://git.launchpad.net/ubuntu/+source/edk2
     source-depth: 1
-    source-commit: b922ebea8da4ddf4a59a652ac570f018672f4872 # import/2025.11-3ubuntu7
+    source-commit: a88d795947662032d58b04d2cb03667a6d87a249 # import/2025.11-3ubuntu8
```
simondeziel: This is from stonking, not resolute. There is 7.1 still baking in `resolute-proposed` that has the same fix though.

snapcraft.yaml
```diff
@@ -890,7 +890,7 @@ parts:
       - spice-server
       - python-runtime
     source: https://git.launchpad.net/ubuntu/+source/qemu
-    source-commit: 96cda2bbb2f530d08f6bda1f0dc186a0e0ce9674 # import/1%10.2.1+ds-1ubuntu3
+    source-commit: 68ab663fefb08c45d883bb4d81fe35a213c1c3b2 # import/1%10.2.1+ds-1ubuntu5
```
simondeziel: This too is from stonking, not resolute. 10.2.1+ds-1ubuntu3.1 is still in `resolute-proposed` AFAIK

snapcraft.yaml
tomponline: 7.1?

tomponline: so we want to stick with [import/2025.11-3ubuntu7](https://git.launchpad.net/ubuntu/+source/edk2/tag/?h=import/2025.11-3ubuntu7) for now because its in resolute?

snapcraft.yaml
tomponline: so we should stick with https://git.launchpad.net/ubuntu/+source/qemu/tag/?h=import/1%2510.2.1%2bds-1ubuntu3 which has the resolute label?

markylaing (approved):

snapcraft.yaml
simondeziel: Sorry, I should have said `2025.11-3ubuntu7.1` is in resolute-proposed. So yeah, we should stick with `2025.11-3ubuntu7` as that's what in resolute.

simondeziel (approved):

PR #1262, opened 2026-06-17
buginfo: Use lxc list --all-projects (5.21-edge)

MggMuggins:
Backport of https://github.com/canonical/lxd-pkg-snap/pull/1247

(cherry picked from commit f31245cceaddb8042a6ab43c1079312094182821)

### DIFF
--- a/snapcraft/commands/buginfo
+++ b/snapcraft/commands/buginfo
@@ -49,7 +49,7 @@ if lxc info >/dev/null 2>&1; then
 
     echo "## Instances"
     echo '```'
-    lxc list
+    lxc list --all-projects
     echo '```'
     echo ""
 


MggMuggins:
Hmm, since this is already present on `latest-edge` and `latest-candidate`, shouldn't the "Branch target" check be smart enough to succeed? Happy to open an issue if so.

simondeziel:
> shouldn't the "Branch target" check be smart enough to succeed? Happy to open an issue if so.

I wouldn't call it smart by any stretch ;)

It wants the target branch name at the end inside parentheses, no backticks and note the `-` instead of `/`.

simondeziel (approved):

simondeziel:
I usually name my branches like `foo-5.21-edge` so the automatic name then requires less mangling to meet the simple check.

tomponline (approved):

PR #1266, opened 2026-06-17
Pin `dqlite`, `lxc` and `lxcfs`

simondeziel:


snapcraft.yaml
```diff
@@ -339,7 +339,7 @@ parts:
 
   dqlite:
     source: https://github.com/canonical/dqlite
-    source-branch: v1.18.x
+    source-commit: f6058ba74838f89b3720c74b7bb4a944bfc36a09 # v1.18.6
```
simondeziel, line 342: This is missing https://github.com/canonical/dqlite/commit/ef0fa37880d0112c19953f75eee45819128a4e99 as it only landed in `main`. The delta between 1.18.6 and `main` is only 3 tiny commits so let me know if you'd prefer to pin to current `main` instead.

tomponline: Can you pin to the release and then cherry pick that fix?

simondeziel: That worked and builds fine:
```
Building dqlite
:: + set -ex
:: + cd ../src
:: + git config user.email noreply@lists.canonical.com
:: + git config user.name 'LXD snap builder'
:: + git cherry-pick -x ef0fa37880d0112c19953f75eee45819128a4e99
:: [detached HEAD 9155079] feat: use MFD_CLOEXEC flag to prevent leak across exec calls (#893)
::  Author: Simon Deziel <simon.deziel@canonical.com>
::  Date: Thu Jun 11 05:58:15 2026 +0000
::  1 file changed, 1 insertion(+), 1 deletion(-)
:: + set +ex
:: + '[' '!' -f ./configure ']'
:: + '[' -f ./autogen.sh ']'
:: + '[' '!' -f ./configure ']'
:: + '[' -f ./bootstrap ']'
:: + '[' '!' -f ./configure ']'
:: + autoreconf --install
```

snapcraft.yaml
```diff
@@ -366,6 +366,8 @@ parts:
       git config user.email "noreply@lists.canonical.com"
       git config user.name "LXD snap builder"
 
+      git cherry-pick -x ef0fa37880d0112c19953f75eee45819128a4e99 # feat: use MFD_CLOEXEC flag to prevent leak across exec calls (#893)
```
tomponline, line 369: we will need source-depth 1 no?

simondeziel: No, 1 means just one level deep so "just the surface" which is a shallow clone.

For reference, in `latest-candidate` we have this:
```
  lxd:
    source: https://github.com/canonical/lxd
    source-commit: 84705553d17aeb8e15032611c321127a06c2f2ff # lxd-6.8
    # XXX: We often cherry-pick for candidate builds so don't do shallow clone (0 means full clone).
    source-depth: 0
```

tomponline (approved):

PR #1272, opened 2026-06-19
lxd-ui: update 0.22 tag (latest-candidate)

edlerd:
tagging to happen at a later time, referencing a commit on main for now (latest-candidate)

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1645,7 +1645,7 @@ parts:
     source: https://github.com/canonical/lxd-ui
     source-depth: 1
     source-type: git
-    source-commit: 0413fc5b563e7c44659cc6e9da24f4eab32cfd00 # 0.22
+    source-commit: 65ff2c18f42dd3852f738f235f8482388de6b974 # 0.22
     plugin: nil
     override-prime: |-
       [ "$(uname -m)" = "riscv64" ] && exit 0


tomponline (approved):

snapcraft.yaml
```diff
@@ -1654,20 +1645,17 @@ parts:
     source: https://github.com/canonical/lxd-ui
     source-depth: 1
     source-type: git
-    source-commit: 0413fc5b563e7c44659cc6e9da24f4eab32cfd00 # 0.22
+    source-commit: 65ff2c18f42dd3852f738f235f8482388de6b974 # 0.22
```
roosterfish, line 1648: @edlerd we landed load balancer pools in LXD for 6.9, but not yet the health checks.

IIRC you mentioned a week or two ago you went ahead and merged the pool+health check support already in the UI gated by the respective API extension? 

Just wanted to check as only pools will be present in 6.9 without health checks.
Likely health checks will get another API extension, it's already the first commit in https://github.com/canonical/lxd/pull/18480.

edlerd: We have to add a guard for the health checks in the UI to not be visible without extension. What is the name of the extension going to be?

roosterfish: I went with `network_load_balancer_pool_health_checks`. The original one was `network_load_balancer_pool`.

edlerd: Adding the UI config guard in https://github.com/canonical/lxd-ui/pull/2079
Also updated screens in [6.9 release tour](https://discourse.ubuntu.com/t/lxd-6-9-has-been-released/84211)

roosterfish: Thanks @edlerd

PR #1273, opened 2026-06-19
lxd-ui: update 0.22 tag, include load balancer health check guard (latest-candidate)

edlerd:
- guard load balancer health checks with api extension


tomponline (approved):

tomponline:
Thanks!

